PR #4, opened 2020-11-16
Import from libraries

kmohrman:


run.py
```diff
@@ -13,25 +13,14 @@
 from coffea import hist, processor
 from coffea.util import load, save
 
+import modules.topeft as topeft
```
kmohrman, line 16: Should we consider changing the name of the modules directory to be something less generic, maybe "topcoffea"?

modules/topeft.py
```diff
@@ -11,13 +11,14 @@
 from coffea.util import load, save
 from optparse import OptionParser
 
+from modules.objects import *
+from modules.corrections import *
```
kmohrman, line 15: It seems that nothing from corrections.py actually ends up getting used here, should we keep it as a placeholder in case we will want something like it in the future, or possibly just consider removing it for now?

run.py
GonzalezFJR: I would leave that name "modules" for the folder containing the python modules with multipurpose functions but wouldn't move the topeft.py file there, as it is not a python module with classes of functions that might be used by other scripts, but an AnalysisProcessor to be executed in coffea. So I would keed the "modules" folder as it was and move that script to another less generic folder (as it was, or in another "analyses" folder if you prefer).

modules/topeft.py
GonzalezFJR: In principle corrections.py will be used for unpacking and storing scale factors into .coffea files, that will be open in topeft but probably won't be needed to be imported.

kmohrman:
Thanks @GonzalezFJR for the feedback on the pull request. We've now restructured the repository as a python package, so we're wondering if you could take another look at the pull request. The updated README describes how to set up the environment and install the package. Please let us know of any comments or questions or suggestions about the changes. Thanks!

GonzalezFJR:
Thanks @kmohrman !!! I'm overall happy with your modifications, but I would keep the modules and plotter folders at their previous level (they're even described in the README!). The "topcoffea" subfolder inside the "topcoffea" repository is redundant, or has it some purpose that I'm missing? On the other hand, I'm not sure what the setup.py script is. Do we need it? If it's somehow useful maybe you can include it in the README with some description (btw thanks for updating the README!!).

kmohrman:
Thanks very much for the feedback @GonzalezFJR! I'll try to address the comments in the order you raised them:

> I would keep the modules and plotter folders at their previous level

The reason why we put these directories under the `topcoffea/topcoffea` directory is so that they would be installed as part of the `topcoffea` package (when we run `pip install .`) so that they can be imported from anywhere in the repository. For example, we can import the functions from the `topcoffea/modules/objects.py` library in the `analysis/topEFT/topeft.py` processor with just `from topcoffea.modules.objects import *`, which we would not be able to do if we had not installed topcoffea as a package. However, I'm not sure if this fully addresses your comment or not, sorry, so please let me know if it does not.

> The "topcoffea" subfolder inside the "topcoffea" repository is redundant, or has it some purpose that I'm missing?

The `topcoffea` subdirectory is supposed to be the main installable package. When looking up python packages, we found that it is apparently somewhat common for this to have the same name as the project [1], but it's of course not required. So if there is a different name you would prefer, please just let us know.

> On the other hand, I'm not sure what the setup.py script is. Do we need it?

The `setup.py` is the script for building/installing the package and is needed for running `pip install .` to install the package. (I should mention that the python package stuff is quite new to me, so I apologize if I'm not using the terminology correctly.)

> If it's somehow useful maybe you can include it in the README with some description

I've added a line to the "Contents" section of the README to describe `setup.py`.

[1] https://betterscientificsoftware.github.io/python-for-hpc/tutorials/python-pypi-packaging/#creating-a-python-package

GonzalezFJR:
Oh, thanks @kmohrman , ok, I see, now I understand! In the version of the main branch I think you don't need to install the package and everything is included in a way that you can include and execute from wherever. At least I had no problems! I'll have a look to the new setup code.

kmohrman:
Thanks @GonzalezFJR! I agree that all of the includes in the current main branch work without any issues. The idea behind the restructuring is aimed towards hopefully making things more flexible. For example, in the current main branch, a file in `analysis/topEFT` couldn't easily import things from `modules` (without manually adjusting the `sys.path`), but with the code set up as a package that can be installed, we can do e.g. `import topcoffea.modules.somelibrary` from anywhere without having to manually adjust the path. Anyway, maybe we can discuss a bit more during the meeting tomorrow.

PR #5, opened 2021-01-14
Moving to HistEFT

GonzalezFJR:
This PR merges the latest updates on HistEFT (including the fix on performance and memory issues) with the code restructuring by Kelci. Now the histograms created in analysis/topEFT/topeft.py include the EFT weight functions and can be evaluated at any WC point.


kmohrman:
Thanks for adding these updates @GonzalezFJR! I tried to test out these changes locally, but ran into a couple issues while trying to get `run.py` to run, so I have just a couple questions. 

First, in `WCFit.py` and in `HistEFT.py` I think the import statements need `topcoffea` before `modules` (e.g. `from topcoffea.modules.WCPoint import WCPoint` instead of `from modules.WCPoint import WCPoint`), as it gave me the error `ModuleNotFoundError: No module named 'modules'` until I made that change to my version. Possibly there is a similar situation in the import statements for `drawSliders.py`.

After making those changes, I tried `run.py` again, and I got a rather lengthy error message, but I think the relevant part is:
```
  File "/afs/crc.nd.edu/user/k/kmohrman/miniconda3/envs/coffea-env-test-xuan-pr/lib/python3.8/site-packages/awkward/array/table.py", line 592, in __getitem__
    return self._contents[where]
KeyError: 'EFTfitCoefficients'
```
So I am not sure if maybe I just have my environment set up incorrectly, or if I am missing some step, or if I am making some other mistake? I am wondering if you could let me know which version of python and coffea you are using, so that I can try to replicate what you are doing in order to determine what is causing my error.

GonzalezFJR:
Thanks for your comments, @kmohrman ! You're right, the code was trying to get a branch with EFT fit coefficients in the samples. I just fixed that: now it includes the fit coefficients only if they exist within the nanoAOD sample, so the code can be run on signal and background samples. I just made a test with a sample without fit coefficients, can you give a try as well? The import errors should be also fixed now.

kmohrman:
Thanks @GonzalezFJR, I tried running it again (on a sample without any EFT fit coefficients), and `run.py` now runs for me without any errors! So I think this PR looks good to me.

PR #22, opened 2021-05-05
Sync-hig: Adding FO objects to objects.py, plus some fix on the fileReader.py

yanfr0818:
1. Adding FO objects to objects.py. Note that those functions are not invoked in the main processor topeft.py at current stage.
2. In topeft.py: varnames['counts'] = np.ones_like(events.MET.pt) -> varnames['counts'] = np.ones_like(events['event']): For some reason, using MET.pt sometimes give error.
3. In fileReader.py: elif 'Runs' in f: -> elif (('Runs' in f) & (not isData)): Prevent datasets using the method 2 when reading the samples.
4. In fileReader.py: nEvents = len(t['MET_pt']) -> nEvents = len(t['event'].array()): Without .array(), nEvents will always be 0.


topcoffea/modules/fileReader.py
```diff
@@ -164,7 +164,7 @@ def GetAllInfoFromFile(fname, treeName = 'Events'):
     f = uproot.open(fname)
     t = f[treeName]
     isData = not 'genWeight' in t#.keys()
-    nEvents = len(t['MET_pt'])
+    nEvents = len(t['event'].array())
```
klannon, line 167: Why are we reading in the whole event as numpy arrays just to count the number of entries?  Uproot's `TTree.num_entries` should do the trick.  (See [here](https://uproot.readthedocs.io/en/latest/uproot.behaviors.TTree.TTree.html#num-entries).)  This is how [coffea does it](https://github.com/CoffeaTeam/coffea/blob/529c36e1695ff44f3d69f243572f9ee4f9bcf551/coffea/processor/executor.py#L1219-L1224).  However, this raises the question: If coffea is going to open all of our files and count their entries, why are we also doing it?

PR #24, opened 2021-05-11
Improve DASsearch Error Handling

### LINKED ISSUE #23: Cannot even import from DASsearch without having dasgoclient in CVMFS
As the title says, importing DASsearch (as samples.py does) will not succeed unless `dasgoclient` is detected at a specific path within CMS.  This is true even if there is no actual attempt to use DAS (e.g. just processing local files).  This makes it impossible to do local development on a laptop.  Two steps should be taken to address this:
1. `dasgoclient` should not be required unless the user actually tries to access DAS.
2. If the user tries to access DAS, it should be sufficient to have `dasgoclient` present anywhere in the system path instead of only at the specific location in CVMFS.  (There is a MacOS version of `dasgoclient` that can be locally installed, after all.)

klannon:
Fixes #23.  Allows the import of DASsearch on systems that don't have `dasgoclient` and only generates errors if an attempt is made to use `dasgoclient`.  Uses `subprocess.run` instead of `os.popen` for more robust and clear error handling.  Looks for `dasgoclient` in path and if it doesn't find it, attempts to add it from `CVMFS`.

### DIFF
--- a/topcoffea/modules/DASsearch.py
+++ b/topcoffea/modules/DASsearch.py
@@ -12,34 +12,50 @@
 
 import os, sys, subprocess
 
-# Check and load voms certificate
-def HasVomsProxy():
-  return True
-  #process = subprocess.Popen('voms-proxy-info', stdout=subprocess.PIPE)
-  #output, error = process.communicate()
-  #output=str(output)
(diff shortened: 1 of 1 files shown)


GonzalezFJR:
Thanks Kevin for dealing with this!

Issue #25, opened 2021-05-11
Finalizing cut and names


bryates:
We talked about this briefly last Friday (5/7), but we need to make sure all the cuts and their corresponding names, e.g. https://github.com/TopEFT/topcoffea/blob/05655b196484c08e20884678df3e5beab6312fb3/analysis/topEFT/topeft.py#L311-L328 and https://github.com/TopEFT/topcoffea/blob/05655b196484c08e20884678df3e5beab6312fb3/analysis/topEFT/topeft.py#L330-L335. We need to make sure all the channels are included (4l is missing) and all the jet/b-jet categories are accounted for.
@GonzalezFJR you said you might try to take a look, but let me know if you're too busy. I can always try updating these.

GonzalezFJR:
I included the 4l categories and categorization according to the sign of the sum of the charge (following TOP-19-001). I also included 2D histograms for njet-nbtag categories, which should be enough to obtain any categorization that we need. 
This is currently in the addingSF branch, in which we're also adding the SFs and systematic categories. We'll open a new PR when some of the ongoing work is finished, probably befor tomorrow's meeting.

bryates:
Thanks! It looks good.

PR #27, opened 2021-05-19
Adding SFs, categories and systematics

GonzalezFJR:
Several updates, mostly in topeft.py
Charge split and 4l categories.
B-tagging SFs.
Uncertainties on plots. Lepton and b-tagging uncertainties.
Implemented JEC uncertainties (but not propagated to the categories, for the moment).


bryates:
Closes #25

kmohrman:
I think the new files in the data directory will not be included in the `topcoffea` package unless you specify them in the `setup.py` file. For example, to include the new `txt` files in `data/JEC/`, you would add `JEC/*.txt` to the `topcoffea` list in the `package_data` dictionary in the `setup.py` file.  

Since the files are not currently listed in the setup script, I think they're probably not being included as part of `topcoffea`. While this is not a problem when running locally with the `-e` option, I think would break if you installed without the `-e`, or if you tried to run with the work-queue version of the code. 

So, for these reasons, I think it might be good to specify those new data files in `setup.py` before merging. To make sure that they are being properly included, you could try uninstalling `topcoffea`, then reinstalling without the `-e` and check that it is able to find the files.

GonzalezFJR:
> I think the new files in the data directory will not be included in the `topcoffea` package unless you specify them in the `setup.py` file. For example, to include the new `txt` files in `data/JEC/`, you would add `JEC/*.txt` to the `topcoffea` list in the `package_data` dictionary in the `setup.py` file.
> 
> Since the files are not currently listed in the setup script, I think they're probably not being included as part of `topcoffea`. While this is not a problem when running locally with the `-e` option, I think would break if you installed without the `-e`, or if you tried to run with the work-queue version of the code.
> 
> So, for these reasons, I think it might be good to specify those new data files in `setup.py` before merging. To make sure that they are being properly included, you could try uninstalling `topcoffea`, then reinstalling without the `-e` and check that it is able to find the files.

Thanks @kmohrman , I added those files to the setup.py script.

bryates:
@GonzalezFJR is there anything left on this PR, or were you just waiting on @kmohrman and I to merge it?

GonzalezFJR:
Yes @bryates I was waiting for you or @kmohrman to merge it when you're happy. I plan to create another PR to introduce the json-based file configs.

kmohrman:
@GonzalezFJR, after the bug fix @bryates mentioned, and the fixes you already made for the `setup.py` script and the `MET_pt` issue, I think this PR looks good to me.

PR #75, opened 2021-08-18
Datacard years

bryates:
This PR removes the year option from `datacard_maker.py` and instead uses the year number in each process name. It also includes all the 2016, 2016 APV, and 2018 background central samples, as well as a fix to the 2017 extension samples.


kmohrman:
Thanks @bryates, this looks good to me. If no one else has any comments, I'd be happy to go ahead and merge it now.

PR #81, opened 2021-08-19
Update variables

bryates:
This PR will update the `ptbl` and `HT` variables. I've found their ranges are too small, and some of the WCs produce large overflows.

I've also updated the README to add instructions for the `EFTFit` tools.


abasnet97:
I was able to use the new changes to make ht and ptbl plots without any issues.

abasnet97:
I noticed a small typo in the README for the job retrieval part. Instead of `fitter.batch1DScanEFT(...)`, it should be `fitter.batchRetrieve1DScansEFT(...)`.

bryates:
> I noticed a small typo in the README for the job retrieval part. Instead of `fitter.batch1DScanEFT(...)`, it should be `fitter.batchRetrieve1DScansEFT(...)`.

Good catch! I pushed the updated README.

bryates:
Removing `idDecayModeNewDMs` allowed me to run over all the signal and background for 2016 (skipping the APV for now), 2017, and 2018.

bryates:
I'm having some issues with this running on earth, so I changed the PR back to a draft. It seems with the larger size of all years + background, the futures multiproces part uses too much of the system resources. I'll look into creating condor jobs instead.

PR #82, opened 2021-08-20
wq: activate fast termination of workers

btovar:
terminate workers that are running tasks for longer than average times the factor fast_terminate_workers.
If this factor is less than 1 or None, it is disabled (default).

It is currently supported with the coffea master branch, but should be generally available with releases after 0.7.6.


klannon:
I have no objections with merging!

Issue #64, opened 2021-07-31
AttachElectronSF() crashing for 2016 APV samples 


kmohrman:
When processing a UL16APV sample on the [EventSelection](https://github.com/TopEFT/topcoffea/tree/EventSelection) branch, the code crashes [here](https://github.com/TopEFT/topcoffea/blob/EventSelection/topcoffea/modules/corrections.py#L137) in `AttachElectronSF()` with the key error pasted below. An example cfg file that should produce the crash is [this](https://gist.github.com/kmohrman/0defd8a030279c9afac11edd77be3ab8). Here is the traceback:

```
Traceback (most recent call last):
  File "/afs/crc.nd.edu/user/k/kmohrman/miniconda3/envs/coffea-env-current/lib/python3.8/concurrent/futures/process.py", line 239, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/afs/crc.nd.edu/user/k/kmohrman/miniconda3/envs/coffea-env-current/lib/python3.8/site-packages/coffea/processor/executor.py", line 186, in __call__
    out = self.function(*args, **kwargs)
  File "/afs/crc.nd.edu/user/k/kmohrman/miniconda3/envs/coffea-env-current/lib/python3.8/site-packages/coffea/processor/executor.py", line 1167, in _work_function
    raise e
  File "/afs/crc.nd.edu/user/k/kmohrman/miniconda3/envs/coffea-env-current/lib/python3.8/site-packages/coffea/processor/executor.py", line 1120, in _work_function
    raise type(e)(str(e) + file_trace).with_traceback(
  File "/afs/crc.nd.edu/user/k/kmohrman/miniconda3/envs/coffea-env-current/lib/python3.8/site-packages/coffea/processor/executor.py", line 1117, in _work_function
    out = processor_instance.process(events)
  File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/topcoffea_EventSelection/topcoffea/analysis/topEFT/topeft.py", line 111, in process
    AttachElectronSF(e_fo,year=year)
  File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/topcoffea_EventSelection/topcoffea/topcoffea/modules/corrections.py", line 137, in AttachElectronSF
    reco_sf          = SFevaluator['ElecRecoSF_{year}'.format(year=year)](eta,pt)
[...]

PR #89, opened 2021-09-01
Adding PU weight

andrea21z:
Adding PU SF and uncertainties.

    - Adding functions to corrections.py to read PU weights.
    - Applying PU SFs in topcoffea.py.


kmohrman:
Thanks @andrea21z for this PR! I agree that it's expected that the yields will change when you implement the PU SFs, so I also agree it's expected that the CI fails. 

Just to double check that the changes associated with the PU SFs are the only changes introduced by this branch, I tried commenting out [this line](https://github.com/TopEFT/topcoffea/blob/PUweight/analysis/topEFT/topeft.py#L289), ran the CI check by hand, and saw agreement between the yields. So I think this indicates there are no unexpected changes to the yields. 

Before we merge this branch, I think the reference yields file the CI compares against will need to be remade. This can be done by just running the [remake_ci_ref_yields.sh](https://github.com/TopEFT/topcoffea/blob/master/analysis/topEFT/remake_ci_ref_yields.sh) script (then of course committing and pushing the updated file).

Additionally, we should also remake this [ref_yields.json](https://github.com/TopEFT/topcoffea/blob/master/analysis/topEFT/test/ref_yields.json) file. These reference yields are not the reference yields that the CI compares against, they are an additional reference that we check by hand by running the [check_yields.sh](https://github.com/TopEFT/topcoffea/blob/master/analysis/topEFT/check_yields.sh) script. This check processes much more data than the CI check does; the bash script is set up to run work-queue, so I can remake and push this file if you want, since it would probably be easiest for someone from ND to do it.

topcoffea/modules/corrections.py
```diff
@@ -194,6 +195,58 @@ def GetBTagSF(eta, pt, flavor, year=2018, sys=0):
...
+## MC histograms from:
+##    https://github.com/CMS-LUMI-POG/PileupTools/
+
+pudirpath = topcoffea_path('data/pileup/')
+
+def GetDataPUname(year='2017', var=0):
+  ''' Returns the name of the file to read pu observed distribution '''
+  if   var== 0: ppxsec = 69200
+  elif var== 1: ppxsec = 72400
+  elif var==-1: ppxsec = 66000
```
bryates, line 216: Two quick questions:

1. Should we hard code these, or maybe just write the formula?
2. These look like 10%. I thought the pp uncertainty was ~5%.

topcoffea/modules/corrections.py
```diff
@@ -132,7 +133,7 @@ def AttachElectronSF(electrons, year=2018):
   # eta = np.abs(electrons.eta)
   eta = electrons.eta
   pt = electrons.pt
-
+  if '2016' in year: year=2016
```
kmohrman, line 136: @andrea21z, the years should be strings, also I think it might be safer to check for an exact matche, e.g.:
```
if year == "2016APV": 
    year = "2016"
```

topcoffea/modules/corrections.py
kmohrman: Another quick comment on this, I think numbers like these would be good to keep in a centralized location, e.g. the [params.json](https://github.com/TopEFT/topcoffea/blob/PUweight/topcoffea/json/params.json) file. You can easily access the numbers in that file using the [get_cut()](https://github.com/TopEFT/topcoffea/blob/695125a6af1c9447de0d165eceb602dbc07f6feb/topcoffea/modules/GetValuesFromJsons.py#L13) function from `GetValuesFromJsons.py`, though we should probably think about renaming that function at some point (since not all of the numbers inside that file are cuts).

andrea21z: 1. Those ppxsec values are introduced in the .root files used to get the PU profile. We left the name they had in the official repository and then: PileupHistogram-goldenJSON-13tev-%s-%sub-99bins.root'%((year), str(ppxsec)

2. You are right, pp uncertainty is ~5% and is what we have:
>>> (72400.-69200)/69200.*100
4.624277456647398
&
>>> (66000.-69200)/69200.*100
-4.624277456647398

bryates: Ok, thanks. So we have to keep the hard coded values for the file names, that's fine. I forgot to think of the uncertainties as shifts, so thanks for pointing out the correct equation!

bryates: Should we still follow @kmohrman's comment about putting these in the `params.json` file, or would you prefer they stay hard coded here since they go into the filename?

bryates: Ok, leaving them here is fine with me. @kmohrman do you have any preference, now that we know they're needed to create the filename string?

kmohrman: I think I'd probably still prefer that they were stored in the param json, if it's not too much trouble.

andrea21z: Ok,  I'll do it. Do you want me to change the function name to, for example, "get_param()"? (instead of "get_cut()")

kmohrman: Sure, if you want do do that, I think that'd be great. Though you'd have to make sure to change it everywhere it gets used.

andrea21z: Done!

bryates: Thanks @andrea21z! I'm running the `check_yields` script on my end now, just to make sure everything works from a fresh pull.

kmohrman: Thanks @andrea21z. @bryates, I'm was actually going to run the same test. I'll can plan to push the results (just to make sure you and I don't both try to push at the same time), unless you would prefer to.

bryates: Sounds good, you can take care of it.

bryates (approved):
Everything looks good to me, and runs on my end.

kmohrman:
I am just going to try running over a UL16APV file, to double check Issue #64 is resolved, then I think we'll be good to go.

kmohrman:
Ok, it ran fine over a 2016 APV file after I fixed a typo in `objects.py`, so I think it's confirmed that this PR fixes #64. I have no objections to merging at this point.

PR #90, opened 2021-09-06
Adding some preliminary tools for plotting 1d quadratic fits

kmohrman:
This branch adds a script ([analysis/topEFT/make_1d_quad_plots.py](https://github.com/TopEFT/topcoffea/blob/quad-fit-tools/analysis/topEFT/make_1d_quad_plots.py)) that uses functions in a new helper file in `modules` ([topcoffea/modules/QuadFitTools.py](https://github.com/TopEFT/topcoffea/blob/quad-fit-tools/topcoffea/modules/QuadFitTools.py)). 
* The `make_1d_quad_plots.py` script is currently more of less a placeholder. It only processes a single example ttH root file, sums the fit arrays, scales the fit to the SM, and creates plots [like these](http://www.crc.nd.edu/~kmohrman/EFT/TopCoffea/testing/example_for_pr_sep06/ttH/) for each 1d fit. It also prints out the values at which the fits scale the cross section up by 10%, since we might eventually need to be able to find information like this (e.g. if we want to create a table similar to TOP-19-001's Table 1). 
* `QuadFitTools.py` contains the functions the `make_1d_quad_plots.py` uses, and is also a bit under-developed. 

These two scripts are pretty basic at this point, so I'm not sure if they're really all that useful to merge in their current state. I just wrote the code because I was becoming concerned about the fact that our only code for getting info from the quadratic fits exists in the fairly dormant c++ based repository [EFTGenReader](https://github.com/TopEFT/EFTGenReader), so I was wondering if we should start thinking about reimplementing some of the tools into `topcoffea` (though if we eventually want to reimplement all the `EFTGenReader` tools in `topcoffea`, we'd probably want to get the `NanoGEN` stuff figured out so we could do the studies at GEN level). Anyway, please let me know if you think it'd be worthwhile to merge this, or if we should hold off till someone has time to develop these tools a bit further. Thanks!


bryates:
Thanks @kmohrman, this looks good! Do we know why some of the plots (e.g. [cQei](http://www.crc.nd.edu/~kmohrman/EFT/TopCoffea/testing/example_for_pr_sep06/ttH/cQei.png)) are flat? Is this just because the script is only opening one nanoAOD file?

kmohrman:
Thanks @bryates. The `cQei` plot is flat because these plots are ttH, and `cQei` is a WC associated with a 2-quark-2-lepton operator (and the other 6 plots that show no dependence are for the other 6 2-quark-2-lepton WCs). Regarding your point about this being only one file, that's true, but even so the statistics are actually not too bad. The file should have around 200k events, which is actually about the same size of sample that we'd usually use for the quadratic plots we'd make using the `EFTGenReader` tools.

bryates:
> Thanks @bryates. The `cQei` plot is flat because these plots are ttH, and `cQei` is a WC associated with a 2-quark-2-lepton operator (and the other 6 plots that show no dependence are for the other 6 2-quark-2-lepton WCs). Regarding your point about this being only one file, that's true, but even so the statistics are actually not too bad. The file should have around 200k events, which is actually about the same size of sample that we'd usually use for the quadratic plots we'd make using the `EFTGenReader` tools.

Thanks! Even after all this time, I still don't fully track in my head which operators go with which samples!

bryates (approved):

PR #91, opened 2021-09-07
work_queue_factory instead of condor_submit_workers

BarrySlyDelgado:
The command work_queue_factory is the proper command to set up workers using the Work Queue Factory. condor_submit_workers does not work with these arguments: -Tcondor -Cfactory.json


klannon:
This seems like a pretty straightforward fix for what looks like an obvious mistake in the docs, so I'd encourage merging this ASAP (@kmohrman and @bryates).  Thanks @BarrySlyDelgado!

PR #92, opened 2021-09-08
Some updates to the yields checks

kmohrman:
In this branch:
* The order we pass the new yields file and reference yield file to `comp_yields.py` has been swapped in both scripts we use for checking yields (i.e. in both [`check_yields.sh`](https://github.com/TopEFT/topcoffea/blob/update-check-yields/analysis/topEFT/check_yields.sh#L26) and the [CI yaml file](https://github.com/TopEFT/topcoffea/blob/update-check-yields/.github/workflows/main.yml#L48)) so that we're now checking `(new yield - ref yield)/(ref yield)` instead of `(ref yield - new yield)/(new yield)`.
* The tolerance for the comparison has been updated from 1e-8 to 1e-5, as suggested by @klannon.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,4 +45,4 @@ jobs:
       - name: Compare yields
         run: |
           cd analysis/topEFT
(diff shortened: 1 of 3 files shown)


bryates (approved):
Looks good to me.

PR #93, opened 2021-09-09
Added fix for large error bars on central samples

jlawrenc:
This should fix the problem I found with large error bars when running over central samples. I have tested this, and this seems to fix the error in all cases I found it.


klannon:
@jlawrenc is too kind.  He's fixed a bug in code I wrote.  That code is completely wrong and @jlawrenc's fix corrects the error, so this is a cut-and-dry bug fix that should be merged ASAP.

PR #95, opened 2021-09-12
Fix bug in Z peak logic

kmohrman:
In this branch, I've attempted to fix what I think is a bug in the way we calculated the "onZ" mask for the 3l categories.
* How the code used to calculate this: We formed pairs of leptons from the whole collection of FO leptons, then checked each pair to see if it was close to the Z peak or not. 
* Why I think we should not do this: The 3l SR category is required to have >=3 FO leptons, and exactly 3 tight leptons (where the 3 tight leptons are required to be the leading 3 leptons). So, when checking if a 3l event contains a pair of leptons close to the Z, I think it seems like we should *only* look at the leading 3 FO leptons (otherwise we could end up with a `True` value for an event where it was e.g. the 1st and 4th leptons that had an invariant mass close to the Z).  
* What this branch changes: We now only check the pairs formed from the leading 3 FO leptons. I also moved the code that finds the mask out of the processor and into a function in `selection.py` ([here](https://github.com/TopEFT/topcoffea/blob/fix-z-mask-logic/topcoffea/modules/selection.py#L340-L346)). Since we'll want a similar mask for one of the `2los` CRs @yanfr0818 is working on, having the code in a function should make this more convenient to implement.  

I also remade the reference yields JSONs. The differences are not large (the largest percent difference in any category is ~3%), which I suppose makes sense since this issue only matters for events that have >3 FO leptons, which there are not all that many of.


topcoffea/modules/selection.py
```diff
@@ -335,3 +335,12 @@ def addLepCatMasks(events):
     events['is_emmm'] = ((n_e_4l==1) & (n_m_4l==3))
     events['is_mmmm'] = ((n_e_4l==0) & (n_m_4l==4))
     events['is_gr4l'] = ((n_e_4l+n_m_4l)>4)
+
+
+# Returns a mask for events with a same flavor opposite sign pair close to the Z
+def get_Z_peak_mask(lep_collection):
+    ll_pairs = ak.combinations(lep_collection, 2, fields=["l0","l1"])
+    zpeak_mask = (abs((ll_pairs.l0+ll_pairs.l1).mass - 91.2)<10.0)
```
bryates, line 343: I believe this is only checking 0 and lepton 1 (unless I'm not thinking of the awkward arrays correctly). Don't we also need to check l0+l2 and l1+l2 for the 3l case?

kmohrman: Yes, we definitely want to form all of the possible pairs, but I think that's what we're doing (at least that's what I'm trying to do anyway). I think the `l0` and `l1` should just be labels for accessing the first and second element in each pair. So for example:
```
import awkward as ak
my_arr = ak.Array([["a","b","c"],["d","e"],["f"],[]])
my_pairs = ak.combinations(my_arr, 2, fields=["l0","l1"])
for pair in my_pairs:
    print(pair)
```
Gives this:
```
[{l0: 'a', l1: 'b'}, {l0: 'a', l1: 'c'}, {l0: 'b', l1: 'c'}]
[{l0: 'd', l1: 'e'}]
[]
[]
```
But I'm not sure if this addresses your concern, or if I'm missing something about the question?

bryates: Thanks for that example. So I was just thinking about awkward arrays incorrectly, and all pairs are considered.

topcoffea/modules/selection.py
```diff
@@ -335,3 +335,12 @@ def addLepCatMasks(events):
     events['is_emmm'] = ((n_e_4l==1) & (n_m_4l==3))
     events['is_mmmm'] = ((n_e_4l==0) & (n_m_4l==4))
     events['is_gr4l'] = ((n_e_4l+n_m_4l)>4)
+
+
+# Returns a mask for events with a same flavor opposite sign pair close to the Z
```
Andrew42, line 340: Maybe expand on this description to say that the mask returns true if _any_ combination of 2 sfos leptons is found in the collection passed to it

Andrew42: you also might want to expand the function to take an extra argument which defines the Z-peak window. This would correspond to what is currently hard coded to `10`. The extra input argument could be called `pt_window` or something.

This would make the function more flexible and also make it clearer what the function is doing when you come across it in `topeft.py`

Andrew42 (changes requested):
I only had two minor changes regarding the new `get_Z_peak_mask()` function. If you or anyone else disagrees with my suggestions then I would be fine with rescinding my change request. Otherwise, I think the PR looks fine to be merged!

topcoffea/modules/selection.py
kmohrman: Thanks @Andrew42, I've made the changes you suggested.

kmohrman:
Thanks again @Andrew42 and @bryates for the comments. Are there any other items to address, or are we ready to merge?

Andrew42 (approved):

PR #97, opened 2021-09-14
Fix xsec

bryates:
This PR fixes all the `xsec = 1` errors. The 2016 `ttH` central sample was also in the background section and folder instead of the signal.


analysis/topEFT/make_jsons.py
```diff
@@ -589,17 +589,17 @@
...
         "path" : "/DYJetsToLL_M-50_TuneCP5_13TeV-amcatnloFXFX-pythia8/RunIISummer19UL18NanoAODv2-106X_upgrade2018_realistic_v15_L1v1-v1/NANOAODSIM",
         "histAxisName": "DY50_centralUL18",
-        "xsecName": "DY50",
+        "xsecName": "DYJetsToLL_M_50_MLM",
     },
     "UL18_singleTop" : {
         "path" : "/ST_s-channel_4f_leptonDecays_TuneCP5_13TeV-amcatnlo-pythia8/RunIISummer20UL18NanoAODv9-106X_upgrade2018_realistic_v16_L1v1-v1/NANOAODSIM",
         "histAxisName": "singleTop_centralUL18",
-        "xsecName": "singleTop",
+        "xsecName": "ST",
```
Andrew42, line 602: would you be opposed to changing this (and elsewhere) to "ST_s-channel"?

topcoffea/cfg/xsec.cfg
```diff
@@ -6,8 +6,14 @@ DYJetsToLL_M_10to50     : 22635.09
 DYJetsToLL_M_50_aMCatNLO: 6025.2
 DYJetsToLL_M_50_MLM     : 6025.2
 
+t                       : 136.02
```
Andrew42, line 9: Which sample does this xsec refer to? Also, I'm not sure how I feel about such a short name, it feels like it could really easily be clobbered accidentally

bryates: That was the single top t-channel. I'll change it to `ST_t-channel`.

analysis/topEFT/make_jsons.py
bryates: Sounds like a good idea.

topcoffea/cfg/xsec.cfg
Andrew42: Wait, but then what does the sample on line 11 refer to?

bryates: That was the s-channel. I see what a mess the names are now, so I'm fixing them.

Andrew42 (approved):
With the name changes I think things are a lot clearer now and I would be fine with having this merged into master, at least from my side.

kmohrman:
Okay, I think we're finally ready to merge this branch. Here's the summary of updates:
* The issues with xsec = 1 have been fixed. If you try to specify an xsec that does not exist, we now raise an exception instead of setting the xsec equal to 1. 
* We've updated all four sets of background samples (UL8, UL17, UL16, and UL16APV) to match the processes used in TOP-19-001. The naming of many of the JSONs and the corresponding `histAxisName` in the JSONs has been updated to improve consistency. The background cfg file has been updated to point to the new JSONs.
* Where there are not UL samples available, we've inserted placeholders with a comment saying `NOTE: PLACEHOLDER` so that we remember to update them later (and here is a GitHub [issue](https://github.com/TopEFT/topcoffea/issues/99) to remind us of this as well). 
* We've updated the cross sections to match the ones used in TOP-19-001, except in cases were we found a newer one to use. However, it is currently unclear what the "official" cross sections are supposed to be, so the cross sections listed in `xsec.cfg` should not be considered to be final at this point. We have initiated a conversation with the TOP conveners to talk about developing a better solution to this common problem.
* As mentioned in the previous point, we will almost certainly want to update the cross sections at some point. However, remaking the whole JSON file can be very time consuming (it takes > 8hours to remake all of the JSONs for all of the background samples). It is thus very tempting to just manually edit the JSONs. However, this is very error prone, so to work around this issue,  I've added a function `replace_xsec_for_dict_of_samples()` that loops through the JSONs and replaces just the cross section value with whatever value is listed in the `xsec.cfg`. This runs in about 1.5 seconds.

Andrew42:
I would suggest adding a reference (where possible) to the xsec.cfg to indicate where the number is coming from

kmohrman:
Thanks @Andrew42. As we discussed on MM, most of the xsec numbers are sort of preliminary placeholders at this point, but I added a few comments to clarify this.

PR #98, opened 2021-09-15
Finalization of nonprompt estimation

sscruz:
First version of the nonprompt estimation. It includes

- Class dataDrivenEstimation.py, takes a dictionary of histograms and integrates the "appl" dimension, keeping only the signal region and building (if requested) the nonprompt estimation from the AR.
- Changes in topeft.py to reweigh also data events by the fake-rate
 - Matching of leptons to MC truth in MC

To-do: 
- Possibility to set a list of processes to be considered in the prompt subtraction
- Smarter handling of empty histograms


analysis/topEFT/run.py
```diff
@@ -83,7 +83,7 @@ def LoadJsonToSampleName(jsonFile, prefix):
   # Read from cfg files
   for f in allInputFiles:
     if not os.path.isfile(f):
-      raise Exception(f'[ERROR] Input file {f} not found!')
+      raise Exception('[ERROR] Input file %s not found!'%f)
```
bryates, line 86: Just curious why you removed the f-string?

sscruz: To be honest, I wasn't aware of the syntax and it was crashing (likely because I wasn't using the right python version when I made the change). But know that you mention it, indeed, this is a valid (and perhaps preferred) way of doing it, I will change it back to what was there before

bryates: Thanks!

analysis/topEFT/dataDrivenEstimation.py
```diff
@@ -0,0 +1,106 @@
...
+        self.outHist=None
+        self.year=year
+
+        self.DDFakes()
+
+    def DDFakes(self):
+        if self.outHist!=None:  # already some processing has been done, so using what is available
+            self.inhist=self.outHist
+            
+        self.outHist={}
```
bryates, line 29: Have you tested this case? You might need to use `self.inhist=self.outHist.copy()` because Python usually uses references, so setting `self.outHist={}` _might_ also clear `self.inhist`.

analysis/topEFT/dataDrivenEstimation.py
```diff
@@ -0,0 +1,106 @@
...
+        for key,histo in self.inhist.items():
+            if key == 'SumOfEFTweights': 
+                # we will deal later with the sum of weights (to do) 
+                continue 
+
+            if not len(histo.values()): # histo is empty, so we just integrate over appl and keep an empty histo
+                self.outHist[key]=histo.integrate('appl')
```
bryates, line 36: Will this run if it is empty? I think it might throw an error, or maybe just a warning.

analysis/topEFT/dataDrivenEstimation.py
```diff
@@ -0,0 +1,106 @@
...
+                continue
+
+            if not self.doDDFakes:
+                # if we are not gonna use the data-driven, then we don't care about the application region, so we get rid of it, and the associated dimension
+                srs=[ histo.integrate('appl',ident) for ident in histo.identifiers('appl') if 'SR' in ident.name]
+                if not len(srs):
+                    continue # to do, this should be a runtime errror
```
bryates, line 43: I agree with the comment. I'll just put this review comment here so we remember to add `raise Exception()`.

analysis/topEFT/dataDrivenEstimation.py
```diff
@@ -0,0 +1,106 @@
...
+                for ident in histo.identifiers('appl'):
+                    hAR=histo.integrate('appl', ident)
+
+                    if 'isAR' not in ident.name:
+                        # if we are in the signal region, we just take the 
+                        # whole histogram integrating out the application region axis
+                        if newhist==None:
+                            newhist=hAR
+                        else:
+                            newhist=newhist+hAR
```
bryates, line 59: Does `+` actually work? I guess `sumw`  is dictionaries. These are only ever SM backgrounds right?

analysis/topEFT/dataDrivenEstimation.py
```diff
@@ -0,0 +1,106 @@
...
+
+                        # we look at data only, and rename it to fakes
+                        hFakes=hAR.group('sample',  hist.Cat('sample','sample'), {'nonprompt' : self.dataName})
+                    
+                        # now we take all the stuff that is not data in the AR to make the prompt subtraction and assign them to nonprompt.
+                        # To do: build a list of samples that we wanna subtract? 
+                        hPromptSub=hAR.group('sample', hist.Cat('sample','sample'), {'nonprompt' : [x.name for x in hAR.identifiers('sample') if x.name != self.dataName]} )
+                        
+                        # now we actually make the subtraction, after normalizing by the luminosity!!
+                        lumi = 1000.0*get_lumi(self.year)
```
bryates, line 72: We should be careful with the years. If we tack this class onto the end of `run.py`, it might give issues when running over multiple years. Take a look at how I handled the years in the datacard maker.
https://github.com/TopEFT/topcoffea/blob/369349eeeb8e5425a5ecfe976e7809ee61878632/analysis/topEFT/datacard_maker.py#L174-L178

topcoffea/modules/corrections.py
```diff
@@ -74,34 +74,36 @@
...
-extLepSF.add_weight_sets(["MuonFR_2016 FR_mva085_mu_data_comb_recorrected %s"%topcoffea_path(basepathFromTTH+'fakerate/fr_2016_recorrected.root')])
-extLepSF.add_weight_sets(["MuonFR_2017 FR_mva085_mu_data_comb_recorrected %s"%topcoffea_path(basepathFromTTH+'fakerate/fr_2017_recorrected.root')])
-extLepSF.add_weight_sets(["MuonFR_2018 FR_mva085_mu_data_comb_recorrected %s"%topcoffea_path(basepathFromTTH+'fakerate/fr_2018_recorrected.root')])
-extLepSF.add_weight_sets(["ElecFR_2016 FR_mva080_el_data_comb_NC_recorrected %s"%topcoffea_path(basepathFromTTH+'fakerate/fr_2016_recorrected.root')])
-extLepSF.add_weight_sets(["ElecFR_2017 FR_mva080_el_data_comb_NC_recorrected %s"%topcoffea_path(basepathFromTTH+'fakerate/fr_2017_recorrected.root')])
-extLepSF.add_weight_sets(["ElecFR_2018 FR_mva080_el_data_comb_NC_recorrected %s"%topcoffea_path(basepathFromTTH+'fakerate/fr_2018_recorrected.root')])
+for year in [2016, 2017, 2018]:
+  for syst in ['','_up','_down','_be1','_be2','_pt1','_pt2']:
+    extLepSF.add_weight_sets([("MuonFR_{year}{syst} FR_mva085_mu_data_comb_recorrected{syst} %s"%topcoffea_path(basepathFromTTH+'fakerate/fr_{year}_recorrected.root')).format(year=year,syst=syst)])
```
bryates, line 79: We _could_ use f-strings here, but the `format` step isn't that complicated, so I won't push for a change.

bryates (changes requested):

analysis/topEFT/dataDrivenEstimation.py
sscruz: I haven't tested this case but an equivalent MWE:
>>> a={'a':'a'}
>>> b=a
>>> a={}
>>> b
{'a': 'a'}

so I think it shouldn't be a problem

analysis/topEFT/dataDrivenEstimation.py
```diff
@@ -0,0 +1,106 @@
+import argparse
+from coffea import hist, processor
+from topcoffea.modules.YieldTools import YieldTools
+from topcoffea.modules.GetValuesFromJsons import get_lumi
+import cloudpickle
+
+class DataDrivenProducer: 
+    def __init__(self, inputHist, outputName, year, doDDFakes=True, doDDFlips=False):
+        self.yt=YieldTools()
+        if type(inputHist) == str: # we are plugging a pickle file
```
Andrew42, line 10: I suggest adding an additional check for the file extension, rather than just relying on the variable being a string, e.g.:
```suggestion
        if isinstance(inputHist,str) and inputHist.endwith('.pkl'): # we are plugging a pickle file
```

analysis/topEFT/dataDrivenEstimation.py
sscruz: yes, it's going to run, and it will return an empty histo. I think this is a desirable output, but I will add a warning

analysis/topEFT/dataDrivenEstimation.py
sscruz: sure, will do

analysis/topEFT/dataDrivenEstimation.py
bryates: Ok. Initially I was afraid about the quadratic fits, but since these are only backgrounds, we don't use `HistEFT`.

analysis/topEFT/dataDrivenEstimation.py
bryates: Ok, thanks.

analysis/topEFT/dataDrivenEstimation.py
bryates: I guess my question then is the `+` operator overloaded for the `hist` class? I've never actually tried this, but if it works, then there's no issue.

analysis/topEFT/dataDrivenEstimation.py
Andrew42: You are correct that python stores references to non-primitive types (basically anything more complicated than a string/int/float), however, you're confusing variable assignment with object modification.

The above code shouldn't result in clearing `self.inhist` b/c what's actually going on is that `self.inhist=self.outHist` is _assigning_ the variable named `self.inhist` to the _object_ that is referred to by `self.outHist`, so now both variables refer to the same object in memory. Now when we later set `self.outHist = {}` that is assigning this variable to an entirely new object (a new instantiated and empty dictionary), however, `self.inhist` still points to the old object in memory.

The case that you're thinking of that you need to be careful is when you assign an object to two different variables and then you _modify_ that object using one of those variables as this will change the actual object in memory and so both variables (which are pointing to the same object) will reflect this change. So for example if `self.outHist={}` was instead `self.outHist.clear()`, that _would_ erase the object that is referred to by `self.inhist` since we're modifying the actual object in memory, rather than just point to a completely new object.

Andrew42 (changes requested):

sscruz:
hi all, 
I've implemented all the points we discussed discussed yesterday. Additionally, I've added the MC truth matching of leptons according to the expected origin (prompt leptons for signal, nonprompt leptons for ttbar, dy, etc). This is done to avoid double counting of evens. For example, if we have a ttH event but one of the leptons happens to be nonprompt this is already taken into account in the data driven estimation, so it has to be removed. Of course this effect is very small in the signal region (and should be even smaller when we change to nanov9 and the electron matching is fixed), but a bit larger in the AR

bryates (approved):
I'm happen with the changes addressing my questions. Thanks @sscruz!

kmohrman:
Thanks @sscruz for this PR! I'm still trying to work my way through it, but in the meantime I have a very minor comment. I'm wondering if the variable called `whatis` could be renamed to something that's a bit more descriptive? Otherwise it's not really clear what this variable is.

sscruz:
> Thanks @sscruz for this PR! I'm still trying to work my way through it, but in the meantime I have a very minor comment. I'm wondering if the variable called `whatis` could be renamed to something that's a bit more descriptive? Otherwise it's not really clear what this variable is.

sure! how about `sampleType`? I'm open to suggestions

kmohrman:
Sure, `sampleType` sounds good to me!

Also, I have another quick question. I'm wondering if it might make sense to put `dataDrivenEstimation.py` inside of `topcoffea/modules`? If I'm understanding correctly, I think the plan would be to use the methods from the `DataDrivenProducer` class in a script (e.g. at the end of the run script), as opposed to running `dataDrivenEstimation.py` from the command line?

sscruz:
perfect, just added the changes. `dataDrivenEstimation.py` supports both being used as a module or from the command line, but I agree it's better in `topcoffea/modules` so I've moved it there

analysis/topEFT/dataDrivenEstimation.py
sscruz: thanks for the suggestion, I've applied it

topcoffea/modules/dataDrivenEstimation.py
```diff
@@ -0,0 +1,161 @@
...
+                    if year not in ['16','17','18']:
+                        raise RuntimeError(f"Sample {sample} does not match the naming convention")
+
+                    if self.dataName == sampleName or self.chargeFlipName == sampleName:
+                        continue # We do not scale data or data-driven at all 
+
+                    for appl in histo.integrate('sample',sample).identifiers('appl'):
+                        if 'AR' not in appl.name: # we only care about AR 
+                            continue
+                        smweight = self.smsow[sample.name] if sample.name in self.smsow else 1 # dont reweight central samples
```
kmohrman, line 80: I'm not sure if I understand this line. Is the purpose to only divide by the sum of weights in the case of an EFT sample using `sample.name in self.smsow` to determine if it's an EFT sample or not? If that's what we're trying to do, I'm wondering if both EFT and central samples will actually be in the `self.smsow` dictionary. The dictionary gets filled inside of this if statement `if 'SumOfEFTweights' in self.inhist`, but I think we make that histogram in all cases (not just in the case of EFT samples). So I'm not sure if this check is doing what it is supposed to do, but I might just be confused about what it's doing.

I think a different way of checking if it's an EFT sample or not (if that's what we're trying to do) might be to just check the number of WCs (using HistEFT's `_nwc`).

kmohrman:
Also, I'm wondering about the `self.dataName='data'` line. We ended up changing the names of the data bins to include the year, so I'm not sure if the `if self.dataName == whatever` lines will do what we want them to do. Maybe we could use `in` instead of checking for exact matches? But I guess it's safer to check for exact matches.

bryates:
> Also, I'm wondering about the `self.dataName='data'` line. We ended up changing the names of the data bins to include the year, so I'm not sure if the `if self.dataName == whatever` lines will do what we want them to do. Maybe we could use `in` instead of checking for exact matches? But I guess it's safer to check for exact matches.

`in` wound definement work. You could also try dynamically generating the list if data years similar to this
https://github.com/TopEFT/topcoffea/blob/db154675cf115eb782d86d75fe71410f28fabb70/analysis/topEFT/datacard_maker.py#L75

sscruz:
hi, I believe the current implementation works and is the safest option. Mind that when I do `self.dataName == sampleName`, `sampleName` is not the name we are plugging in coffea but really the portion of the name that refers to the sample itself. This is done using regular expressions. For example, see 
https://github.com/TopEFT/topcoffea/pull/98/files#diff-25cf1ff6f1e89509998f31f541dfa2834c1d7419de402157663e63f61f1655eaR106-R109

kmohrman:
> hi, I believe the current implementation works and is the safest option. Mind that when I do `self.dataName == sampleName`, `sampleName` is not the name we are plugging in coffea but really the portion of the name that refers to the sample itself. This is done using regular expressions. For example, see
> https://github.com/TopEFT/topcoffea/pull/98/files#diff-25cf1ff6f1e89509998f31f541dfa2834c1d7419de402157663e63f61f1655eaR106-R109

Thanks for explaining this @sscruz. But I think I'm still a bit confused, sorry, because when I print out the `sampleName`, for the data sample `data_UL17` the `sampleName` seems to be `data_`, which will not match `data`.

sscruz:
I see, then the problem is that we are not following the exact naming convention for data and MC samples in the `json`: for MC (for example in https://github.com/TopEFT/topcoffea/blob/master/topcoffea/json/signal_samples/central_UL/UL16_ttHJetToNonbb_M125.json)  we have `{datasetname}UL16`, while for data we have `{datasetname}_UL16`. Hence the regexp is interpreting `data_`. I see two options: to change the naming in the data json so it matches data or just change the hardcoded value `self.dataName' to `data_` (easiest but a bit ugly :) )

kmohrman:
Ah, okay, that makes sense. It's easy to remake the JSONs for the data samples, so I can just update the naming and remake them right now (I guess I'll have to make a new branch for that though, as I probably cannot push to your fork).

topcoffea/modules/dataDrivenEstimation.py
```diff
@@ -0,0 +1,161 @@
...
+                        hPromptSub=hAR.group('sample', hist.Cat('sample','sample'), newNameDictNoData )
+                        
+                        # now we actually make the subtraction
+                        hPromptSub.scale(-1)
+                        hFakes=hFakes+hPromptSub
+
+                        #scale back by 1/lumi as if it were a MC, so it can be used transparently downstream
+                        scaleDict={}
+                        for name, year in addedNonPrompts:
+                            scaleDict[name]=1/(1000*get_lumi(year))
```
kmohrman, line 129: @sscruz, the code is crashing here for me, I think because `year` is `17` while the key in the lumi json is `2017`.

kmohrman:
@sscruz, I think there might be two potential issues happening in [this line](https://github.com/sscruz/topcoffea/blob/NonPromptEstimation/topcoffea/modules/dataDrivenEstimation.py#L80):
1. I think the first issue is that *all* samples (not just EFT samples) will be present in the `self.smsow` dictionary. I think this is because all samples in the histogram are looped over when filling the dictionary [here](https://github.com/sscruz/topcoffea/blob/NonPromptEstimation/topcoffea/modules/dataDrivenEstimation.py#L36). The dictionary is filled inside of the if statement `if 'SumOfEFTweights' in self.inhist`, but all pkl files will have a  `SumOfEFTweights` histogram, not just pkl files of EFT samples. I'm not sure of the best way to solve this. Maybe when filling the dictionary, we could check the number of WCs with HistEFT's `_nwc`, i.e. check SumOfEFTweights._nwc, and fill the dictionary only if it's >0.   
1. I think the second issue is that, regardless of whether or not a sample is in the `self.smsow` dictionary, the `sample.name in self.smsow` check will be `False`. This seems to be due to the fact that the keys of the `self.smsow` are not actually strings, they are a `StringBin` object. So, the `sample.name` will be something like `tHq_privateUL17`, but the corresponding key in the dictionary is `<StringBin (tHq_privateUL17) instance at 0x7fe7e2f0a7f0>`. To solve this second issue, I think that when the dictionary is filled [here](https://github.com/sscruz/topcoffea/blob/NonPromptEstimation/topcoffea/modules/dataDrivenEstimation.py#L36), we could just just fill the dictionary like this:
    ```
    self.smsow = {}
    if SumOfEFTweights._nwc > 0:
        for identifier in SumOfEFTweights.identifiers('sample'):
            self.smsow[identifier.name] = SumOfEFTweights.integrate('sample', identifier).values()[()][0]

    ``` 
    With this method, the keys will be the strings, not the  `StringBin` objects.

sscruz:
hi @kmohrman 
thanks! you are right, I've just pushed a fix

kmohrman:
Thanks @sscruz! I have just a couple other questions, sorry. On [this line](https://github.com/sscruz/topcoffea/blob/NonPromptEstimation/topcoffea/modules/dataDrivenEstimation.py#L82), I'm not sure the `scale` is working. I printed out the values before and after scaling, and they did not change. I think this is because the keys of the `scale_dict` do not mach the axes of the `histo`. Specifically, I think we're not specifying the `systematic` axis. I think we could work around this by replacing [this line](https://github.com/sscruz/topcoffea/blob/NonPromptEstimation/topcoffea/modules/dataDrivenEstimation.py#L81) with this:
```
scale_dict[(sample,)] = 1000*get_lumi('20'+year)/smweight
``` 
Then scaling the hist like this `histo.scale( scale_dict, axis=('sample',))` (following the example [here](https://coffeateam.github.io/coffea/api/coffea.hist.Hist.html#coffea.hist.Hist.scale)). This seems a bit simpler anyway, since I think we really only need a different element in the dictionary for each sample, not for each combination of sample and application region. 

Also, I think there's one other potential issue with this scaling. It seems we're also scaling the data samples, since we fill the dictionary inside of a loop that's looping over `histo.identifiers('sample')`. Maybe we could add an if statement before we fill the dictionary, to `continue` if the sample name has `data` in it?

sscruz:
thanks! I'm checking the first point. About the second, point, I'm a bit confused, isn't that what we are doing here https://github.com/TopEFT/topcoffea/pull/98/files#diff-25cf1ff6f1e89509998f31f541dfa2834c1d7419de402157663e63f61f1655eaR74-R75 ? Or am I missing something?

kmohrman:
About the second point, yes, sorry, you're right. That was a mistake on my end. I was running over an old pkl file that still had the extra underscore in the data sample names, so the if statement you pointed to was not catching the data samples, so all of the data samples ended up in my scale dictionary.

sscruz:
hi, just a small update. After discussing with @kmohrman we realized that the prompt subtraction won't work when using EFT samples, because one cannot sum hists with EFT hists. So we have agreed to 
1) implement the functionality that allows to do this (@kmohrman is working on this, I can help if needed)
2) implement a list of processes to be considered for the subtraction, so we are safe in the case we want to run both over the official and private samples. I just pushed this in the previous commit

Andrew42 (approved):

PR #104, opened 2021-09-28
Implement golden json mask

kmohrman:
In this branch:
* Implemented a mask to make sure that we only use data from good runs (defined in the "golden JSONs") using `coffea.lumi_tools.LumiMask`. This mask is only applied when processing a data sample.  
* Copied the "golden json" files (which are apparently actually `.txt` files, not `.json` files...) to a new directory in `topcoffea/topcoffea/data/goldenJsons/`, and added a [README](https://github.com/TopEFT/topcoffea/tree/golden-json-mask/topcoffea/data/goldenJsons) to that directory to document the relevant [twiki](https://twiki.cern.ch/twiki/bin/view/CMS/TWikiLUM).
* Updated the luminosity values in the [lumi.json](https://github.com/TopEFT/topcoffea/blob/golden-json-mask/topcoffea/json/lumi.json) file to match the values listed in the twiki mentioned above.
* Remade the reference yield files (since the change to `lumi.json` modified the yields by about half a percent). 

Thank you to @rgoldouz for bringing this issue to our attention.


bryates (commented):
Thanks @kmohrman! Is there any official word on why sonly 2017 has UL in the name? I suppose these won't change between versions. Lost data is lost data.

kmohrman:
I'm not sure on the naming. According to the twiki linked above, the integrated luminosity does change slightly from the pre-legacy to legacy json. We're currently using the files that correspond to "Golden JSON (Legacy)" on the twiki. I copied the files from the `Legacy` directory (e.g. `/afs/cern.ch/cms/CAF/CMSCOMM/COMM_DQM/certification/Collisions16/13TeV/Legacy_2016/`, and similar for 2017 and 2018) in all three cases. Comparing the txt files against the files linked from the twiki, each matches the "Golden JSON (Legacy)" one for its year. But I'm not sure why the txt files themselves are named slightly differently, with the 2017 one being the only one with UL in its name.

bryates:
> I'm not sure on the naming. According to the twiki linked above, the integrated luminosity does change slightly from the pre-legacy to legacy json. We're currently using the files that correspond to "Golden JSON (Legacy)" on the twiki. I copied the files from the `Legacy` directory (e.g. `/afs/cern.ch/cms/CAF/CMSCOMM/COMM_DQM/certification/Collisions16/13TeV/Legacy_2016/`, and similar for 2017 and 2018) in all three cases. Comparing the txt files against the files linked from the twiki, each matches the "Golden JSON (Legacy)" one for its year. But I'm not sure why the txt files themselves are named slightly differently, with the 2017 one being the only one with UL in its name.

Ok thanks. I think we can merge this PR now, but maybe we can keep an eye on this for the future, and open another PR if they do ever change the names/content.

bryates (approved):

kmohrman:
I agree. I've created an issue (#105) so we remember to keep an eye on it as you suggested.

PR #106, opened 2021-09-28
Update ref yields

kmohrman:
This branch updates the reference yields to include the effects of the PU SFs. (In the previous PR, the ref yield file was updated [here](https://github.com/TopEFT/topcoffea/commit/ac2d454b472a01688ec9a86141afc8d3521ca3b5), before [this commit](https://github.com/TopEFT/topcoffea/commit/ba9d29ce8df4713488e46764aaea373bd129d6d4) that added in the PU SFs.)

As a minor and unrelated update, this branch also adds in the 3l AR bin in the `appl` category in the 3l CR histogram (which we forgot to include when defining the CRs).


bryates (approved):
Looks good to me. I ran `check_yields.sh` on my end, and the only thing not below `1e-16` is 
```
tllq
    3l_p_offZ_2b
        1.5701836219540973e-05 -> NOTE: This is larger than tolerance (1e-05)!
```
which I believe we've seen before.

kmohrman:
Thanks for checking! And yes, I think that is the same super weird, seemingly inexplicable 1e-5 issue that came up like a month ago...

PR #108, opened 2021-10-01
Add eft and noneft bins

kmohrman:
This branch adds the ability to combine the `sumw` and `sumw2` values in EFT and non-EFT bins. This is necessary for the prompt subtraction for the data driven nonprompt background estimation (where we need to combine the EFT signal samples with the SM background samples, and subtract the result from the data samples). The basic idea is that when combining an EFT and a non-EFT bin, the non EFT bin value should be added to the 0th element of the EFT bin's array (i.e. the non-EFT bin's value should be combined with the SM part of the EFT bin's array).

This branch also implements the ability to combine bins with regular errors to bins with errors of `None` (which we have when processing EFT samples with `do_errors` turned off). There are many ways this could be implemented, but the approach we went with is that "`None` plus anything equals `None`".


bryates (commented):
Thanks @kmohrman, this looks good to me. I think @sscruz's fake factor work was the main reason we needed this, so I'll let him run any tests needed to confirm this all works.

kmohrman:
> Thanks @kmohrman, this looks good to me. I think @sscruz's fake factor work was the main reason we needed this, so I'll let him run any tests needed to confirm this all works.

Thanks for the comments @bryates. I tried to double check this again by doing the calculations that the `dataDrivenEstimatro.py` script does by hand. I printed out the values in a particular bin of one of the CR categories at the beginning of `dataDrivenEstimatro.py`. If everything (both the `dataDrivenEstimatro.py` script and the `add()` and `sum()` methods of `HistEFT` that it uses) are working correctly, then the value in the "nonprompt" bin it adds to the samples axis should be equal to the following (sorry, it's hard to read):

```
value in new nonprompt bin = ((value of data in AR bin) - ((value of signal AR bin)/(sum of weights at SM) + (value of background AR bin))(lumi for appropriate year))/(lumi for appropriate year)
```
Calculating that by hand from the the input histogram, the result agreed really well with the value the code dumped into the nonprompt bin (the numbers were off by a percent difference around 1e-8, but I assume that's probably just rounding issues).

So, at least from my end, I have not been able to find any indications that the modifications to `add()` and `sum()` are causing the `dataDrivenEstimatro.py` to calculate the non prompt contribution incorrectly.

I'm wondering if there were any additional checks or tests you'd have in mind before we proceed with merging the branch?

kmohrman:
I'm wondering if there any objections to this branch and/or any tests that @bryates or @sscruz would like to run? Otherwise, should we try to get it merged?

bryates:
> I'm wondering if there any objections to this branch and/or any tests that @bryates or @sscruz would like to run? Otherwise, should we try to get it merged?

I have no objections.

PR #109, opened 2021-10-04
Fix crash when processing data, and fix normalization bug

kmohrman:
This branch is supposed to fix two bugs:
* After merging PR #98, the code was crashing [here](https://github.com/TopEFT/topcoffea/blob/master/analysis/topEFT/topeft.py#L315) when processing data. Even though we do not use the PU SFs for data, the code was still trying to calculate them, which would be fine (since they do not actually get applied), except for the fact that it was crashing in the process. To work around this, I just put the PU SF calculation inside an `if not isData` statement. 
* Also, as @bryates pointed out last week, the issue with trying to use the length of the WC list to determine if we're dealing with an EFT sample or not also showed up in the processor. That check does not work when processing eft and non-EFT samples together, since all histograms will have a WC list that is the superset of all WCs in all of the samples being processed. As a quick fix, I just swapped out that check with an `eft_coeffs is not None` check, as we already set the `eft_coeffs` variable [here](https://github.com/TopEFT/topcoffea/blob/master/analysis/topEFT/topeft.py#L333), so it seems like we might as well use it.

### DIFF
--- a/analysis/topEFT/run.py
+++ b/analysis/topEFT/run.py
@@ -161,7 +161,7 @@ if __name__ == '__main__':
 
   # Run the processor and get the output
   tstart = time.time()
-  output = processor.run_uproot_job(flist, treename=treename, processor_instance=processor_instance, executor=processor.futures_executor, executor_args={"schema": NanoAODSchema,'workers': nworkers, 'pre_workers': 1}, chunksize=chunksize, maxchunks=nchunks)
+  output = processor.run_uproot_job(flist, treename=treename, processor_instance=processor_instance, executor=processor.futures_executor, executor_args={"schema": NanoAODSchema,'workers': nworkers}, chunksize=chunksize, maxchunks=nchunks)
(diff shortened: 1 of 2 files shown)


kmohrman:
Also updated `run.py` to remove `pre_workers` from the executor args. I'm not actually sure what this argument was supposed to do or why we needed it, but it's been causing the CI to crash ever since coffea 0.7.8 was released a couple days ago (since that argument is apparently no longer recognized), so I tried removing it, and that seems to fix the issue. 

But if anyone knows what this argument was used for, or whether or not it's important, please let me know.

bryates (commented):
Thanks @kmohrman, `eft_coeffs is not None` fix seems more elegant than the one I was working on. I'm running some tests on my end to make sure it all works correctly in the datacard maker.

kmohrman:
Thanks for checking! I'll go ahead and merge it now.

PR #111, opened 2021-10-04
JSONs with skims

sscruz:
I'm committing here the json file corresponding to the skims. The skims require at least two loose leptons. For data, events are required to fire the analysis level triggers and the dataset overlap is removed. The code used to run the skims, generate the cfg and json files, and to check their integrity can be found here: 
https://github.com/sscruz/nanoAOD-tools/tree/GlobalEFTskimmer


analysis/topEFT/topeft.py
```diff
@@ -362,9 +365,10 @@ def process(self, events):
         bmask_atleast2med = (nbtagsm>=2) # Used for 3l SR
 
         # Charge masks
-        charge2l_p = ak.fill_none(((l0.charge+l1.charge)>0),False)
-        charge2l_m = ak.fill_none(((l0.charge+l1.charge)<0),False)
+        charge2l_p = ak.fill_none(((l0.charge)>0),False)
+        charge2l_m = ak.fill_none(((l0.charge)<0),False)
         charge2l_0 = ak.fill_none(((l0.charge+l1.charge)==0),False)
+        charge2l_pm = ak.fill_none(((l0.charge+l1.charge)!=0),False)
```
bryates, line 371: We should think of a better name. To me `pm` is plus/minus, meaning opposite sign.

sscruz: thanks! I just pushed the change

sscruz:
hi, to overcome that issue, we can rename `2lss_p` to `2l_l0p` or similar. I'd avoid changing the code logic in that way because 
1) we would have to enforce by hand the fact that the SS signal region and the OS application region have the selection and binning
2) we have to fully reimplement the charge flip code in the `dataDrivenEstimation.py`, and do it differently from the way we implement nonprompts, which is a bit odd as well...

topcoffea/modules/selection.py
```diff
@@ -175,6 +174,7 @@ def trgPassNoOverlap(events,is_data,dataset,year):
 
     # In case of data, check if events overlap with other datasets
     if is_data:
+        trg_passes = passesTrgInLst(events,dataset_dict[year][dataset])
```
Andrew42, line 177: @sscruz thanks for catching this!

Just to be explicit about what the issue is so that I'm not missing something. The reason for this logic change is for the edge case where an event from dataset "A" doesn't pass any of the analysis triggers for dataset "A", but _does_ pass an analysis trigger from dataset "B". This means that it could be double counted b/c the overlap removal won't remove the event when it showed up in dataset "A" (b/c it didn't pass any of the analysis level triggers for that dataset) while still allowing the event to be included when it shows up in dataset "B". Is that correct?

I'm pretty sure this should be a pretty small effect. When I tested the original implementation by shuffling the order of the overlap removal, I saw _almost_ no differences in the distributions. Since the differences were so small I didn't investigate it any further at the time, but I'm guessing this is the reason for those differences.

In any case, this is a good catch and I am fine with this change!

topcoffea/modules/corrections.py
```diff
@@ -89,17 +93,23 @@
 def AttachPerLeptonFR(leps, flavor, year=2018):
   if year == '2016APV': year = 2016
   for syst in ffSysts:
-    fr=SFevaluator['{flavor}FR_{year}{syst}'.format(flavor=flavor,year=year,syst=syst)](np.abs(leps.eta), leps.conept )
+    fr=SFevaluator['{flavor}FR_{year}{syst}'.format(flavor=flavor,year=year,syst=syst)](leps.conept, np.abs(leps.eta) )
     leps['fakefactor%s'%syst]=ak.fill_none(-fr/(1-fr),0) # this is the factor that actually enters the expressions
-  fr=SFevaluator['{flavor}FR_{year}'.format(flavor=flavor,year=year)](np.abs(leps.eta), leps.conept )
+  fr=SFevaluator['{flavor}FR_{year}'.format(flavor=flavor,year=year)](leps.conept, np.abs(leps.eta) )
```
kmohrman, line 98: I have a quick question about this line. I'm not sure if I'm just missing something here, but it looks like this line is already covered by the for loop, since the loop includes `''` as one of the strings to loop over. If that's the case, I'm wondering if we could just remove the line?

topcoffea/modules/dataDrivenEstimation.py
```diff
@@ -94,8 +93,7 @@ def DDFakes(self):
                     else:
                         if "isAR_2lSS_OS"==ident.name:
                             if key =='njets':
```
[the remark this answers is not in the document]
kmohrman: I think it's not needed. Also, I just tried pulling the changes and running, and it's unhappy about the indentation, but I think that can be resolved by just removing the lines.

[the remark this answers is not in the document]
bryates: Thanks.

bryates:
> Weird, when I ran it by hand on my branch, I got:
> 
> ```
> FileNotFoundError: [Errno 2] No such file or directory: 'NAOD-00000_18449.root'
> ```

Ah yes, I did move things around, so it has to be run from the main topcoffea folder, not `analysis/topEFT` anymore. If you just run `pytest` form the main folder, it should _hopefully_ run. I'm trying that now.

bryates:
> > Weird, when I ran it by hand on my branch, I got:
> > ```
> > FileNotFoundError: [Errno 2] No such file or directory: 'NAOD-00000_18449.root'
> > ```
> 
> Ah yes, I did move things around, so it has to be run from the main topcoffea folder, not `analysis/topEFT` anymore. If you just run `pytest` form the main folder, it should _hopefully_ run. I'm trying that now.

This is the culprit:
https://github.com/TopEFT/topcoffea/blob/843dc07f8171f52a973d33468e738d2900c9a824/analysis/topEFT/topeft.py#L426
since `isSR_2l` changed to `isSR_2lSS`, I'll need to update the datacard maker on this branch. I'll do that now.

kmohrman:
@bryates, any thoughts on what might be causing the codecov issue?

PR #116, opened 2021-10-15
Added a README for pytest

bryates:
This PR adds a README file to the `tests` folder, detailing the various `pytest` scripts.

### DIFF
--- a/tests/test_make_1d_quad_plots.py
+++ b/tests/test_make_1d_quad_plots.py
@@ -8,7 +8,7 @@ def test_make_1d_quad_plots():
         "-i",
         "NAOD-00000_18449.root",
         "-r",
-        "/afs/crc.nd.edu/user/b/byates2/topcoffea/"
+        "./"
     ]
 
     # Run make_1d_quad


bryates:
> I noticed this warning in the CI for the quad fit plotting script https://github.com/TopEFT/topcoffea/pull/116/checks?check_run_id=3908512841#step:8:594-597 It doesn't cause an error, but might be worth looking into.

I wonder if this is `nan` is the issue
`ctli crosses 1.1 at: [nan, nan]`

kmohrman:
@bryates, the PU reweighting (or anything else that happens after we make the NAOD files) could not have caused this issue. That script runs directly on the NAOD file itself. It is not currently used for anything, I just wrote it extemporaneously some weekend a couple months ago because I realized that we had no tools in this repository for extracting information about the quadratic fits, and I thought that if/when we want to be able to do that (e.g. if we're asked to include a "Leading processed affects" column in this analysis' version of TOP-19-001's Table 1), it'd be nice to be able to do that with python based tools instead of having to go back to our old c++ tools. So, that script was just supposed to be a sort of example we could build onto later on.

Regarding the actual warning you're pointing to, I think it's probably not surprising that the code is having a hard time figuring out where the quadratic fit for ctli crosses 1.1, since I think we're testing with a ttH file. That WC is associated with one of the operators involving two quarks and two lentos, and has no effect on the ttH.

bryates:
> @bryates, the PU reweighting (or anything else that happens after we make the NAOD files) could not have caused this issue. That script runs directly on the NAOD file itself. It is not currently used for anything, I just wrote it extemporaneously some weekend a couple months ago because I realized that we had no tools in this repository for extracting information about the quadratic fits, and I thought that if/when we want to be able to do that (e.g. if we're asked to include a "Leading processed affects" column in this analysis' version of TOP-19-001's Table 1), it'd be nice to be able to do that with python based tools instead of having to go back to our old c++ tools. So, that script was just supposed to be a sort of example we could build onto later on.
> 
> Regarding the actual warning you're pointing to, I think it's probably not surprising that the code is having a hard time figuring out where the quadratic fit for ctli crosses 1.1, since I think we're testing with a ttH file. That WC is associated with one of the operators involving two quarks and two leptons, and has no effect on the ttH.

Ok, that's a good point.

PR #121, opened 2021-11-03
Adding Rochester corrections and L1 prefiring weights

andrea21z:
This PR includes:
 - Rochester corrections
 - L1 trigger corrections (only for nanoAODv9)
 - Fixing problem with btagging SFs
 - Improving functions in `corrections.py` file
 
The CI check is failing because yields change wrt. the ref_yields.json. I have updated the new_yields json file running `remake_ci_ref_yields.sh`, as in the previous PR, but it doesn't work.


kmohrman:
@andrea21z, thanks very much for this PR! Overall I think it looks really good. Regarding your question about the CI, I think it looks like you updated the reference yields correctly, and that part of the CI seems to be passing. The part that's failing looks to be the data card comparisons. If it would be helpful, I can try to update those now. We'll also need to update the other set of reference yields (not the one the CI checks, but the bigger comparison that we run with work-queue). Since that requires running WQ, it'd probably be easiest if someone with an ND account updated that ref file, so I can try to update that this morning as well.

But other than updating those files, I think it looks good to me. However, I've only had time to take just a quick look so far, so I'll plan to spend some time looking through the rest of the PR more closely, and will let you know if I run into any questions.

Thanks!

analysis/topEFT/topeft.py
```diff
@@ -15,7 +15,7 @@
 
 from topcoffea.modules.GetValuesFromJsons import get_param
 from topcoffea.modules.objects import *
-from topcoffea.modules.corrections import SFevaluator, GetBTagSF, jet_factory, GetBtagEff, AttachMuonSF, AttachElectronSF, AttachPerLeptonFR, GetPUSF
+from topcoffea.modules.corrections import SFevaluator, GetBTagSF, jet_factory, GetBtagEff, AttachMuonSF, AttachElectronSF, AttachPerLeptonFR, GetPUSF, apply_roccor
```
bryates, line 18: We should stick to a convention, something like `ApplyRochesterCorrections` (I guess `jet_factory` slipped passed us!), or even put the corrections inside `SFevaluator`.

analysis/topEFT/topeft.py
```diff
@@ -51,7 +51,7 @@ def construct_cat_name(chan_str,njet_str=None,flav_str=None):
 
 class AnalysisProcessor(processor.ProcessorABC):
 
-    def __init__(self, samples, wc_names_lst=[], hist_lst=None, do_errors=False, do_systematics=False, split_by_lepton_flavor=False, skip_signal_regions=False, skip_control_regions=False, dtype=np.float32):
+    def __init__(self, samples, wc_names_lst=[], hist_lst=None, do_errors=False, do_systematics=False, split_by_lepton_flavor=False, skip_signal_regions=False, skip_control_regions=False, dtype=np.float32, muonSyst='nominal'):
```
bryates, line 54: Not a huge deal, but I personally prefer `dtype` to be the last argument.

analysis/topEFT/topeft.py
```diff
@@ -308,8 +316,8 @@ def process(self, events):
             pt = goodJets.pt; abseta = np.abs(goodJets.eta); flav = goodJets.hadronFlavour
 
             bJetSF   = GetBTagSF(abseta, pt, flav, year)
-            bJetSFUp = GetBTagSF(abseta, pt, flav, year, sys=1)
-            bJetSFDo = GetBTagSF(abseta, pt, flav, year, sys=-1)
+            bJetSFUp = GetBTagSF(abseta, pt, flav, year,sys=1)
+            bJetSFDo = GetBTagSF(abseta, pt, flav, year,sys=-1)
```
bryates, line 320: Can we put the space back in between `year` and `sys`?

analysis/topEFT/topeft.py
```diff
@@ -318,12 +326,8 @@ def process(self, events):
             pMC     = ak.prod(bJetEff       [isBtagJetsMedium], axis=-1) * ak.prod((1-bJetEff       [isNotBtagJetsMedium]), axis=-1)
             pData   = ak.prod(bJetEff_data  [isBtagJetsMedium], axis=-1) * ak.prod((1-bJetEff_data  [isNotBtagJetsMedium]), axis=-1)
             pDataUp = ak.prod(bJetEff_dataUp[isBtagJetsMedium], axis=-1) * ak.prod((1-bJetEff_dataUp[isNotBtagJetsMedium]), axis=-1)
-            pDataDo = ak.prod(bJetEff_dataDo[isBtagJetsMedium], axis=-1) * ak.prod((1-bJetEff_dataDo[isNotBtagJetsMedium]), axis=-1)
-
+            pDataDo = ak.prod(bJetEff_dataDo[isBtagJetsMedium], axis=-1) * ak.prod((1-bJetEff_dataDo[isNotBtagJetsMedium]), axis=-1)           
             pMC      = ak.where(pMC==0,1,pMC) # removeing zeroes from denominator...
-            btagSF   = pData  /pMC
-            btagSFUp = pDataUp/pMC
-            btagSFDo = pDataDo/pMC
```
bryates, line 326: Were these lines removed on accident?

analysis/topEFT/topeft.py
```diff
@@ -342,20 +346,21 @@ def process(self, events):
                 weights_dict[ch_name].add("btagSF",pData/pMC,pDataUp/pMC,pDataDo/pMC)
                 # Trying to calculate PU SFs for data causes a crash, and we don't apply this for data anyway, so just skip it in the case of data
                 weights_dict[ch_name].add('PU', GetPUSF((events.Pileup.nTrueInt), year), GetPUSF(events.Pileup.nTrueInt, year, 1), GetPUSF(events.Pileup.nTrueInt, year, -1))
-            if "2l" in ch_name:
-                weights_dict[ch_name].add("lepSF", events.sf_2l, events.sf_2l_hi, events.sf_2l_lo)
-                weights_dict[ch_name].add("FF"   , events.fakefactor_2l, events.fakefactor_2l_up, events.fakefactor_2l_down )
-                if isData:
-                    weights_dict[ch_name].add("fliprate"   , events.flipfactor_2l)
```
bryates, line 349: Looks like these lines were removed.

analysis/topEFT/topeft.py
```diff
@@ -558,7 +563,7 @@ def process(self, events):
 
                 # In the case of "nominal", or the jet energy systematics, no weight systematic variation is used (weight_fluct=None)
                 weight_fluct = syst
-                if syst in ["nominal","JERUp","JERDown","JESUp","JESDown"]: weight_fluct = None # No weight systematic for these variations
+                if syst in ["nominal","JERUp","JERDown","JESUp","JESDown","MuonESup","MuonESdo"]: weight_fluct = None # No weight systematic for these variations
```
bryates, line 566: `MuonESdo` -> `MuonESdown`

topcoffea/modules/corrections.py
```diff
@@ -285,3 +289,38 @@ def GetPUSF(nTrueInt, year, var=0):
...
+        errors[hasgen_flat] = np.array(ak.flatten(errspread))
+        errors[~hasgen_flat] = np.array(ak.flatten(errsmear))
+        corrections = ak.unflatten(corrections, ak.num(mu.pt, axis=1))
+        errors = ak.unflatten(errors, ak.num(mu.pt, axis=1))
+    else:
+        corrections = rochester.kScaleDT(mu.charge, mu.pt, mu.eta, mu.phi)
+        errors = rochester.kScaleDTerror(mu.charge, mu.pt, mu.eta, mu.phi)
+    if var==0: return(mu.pt * corrections) #nominal
+    elif var==1: return(mu.pt * corrections + mu.pt * errors) #up 
+    elif var==-1: return(mu.pt * corrections - mu.pt * errors) #down
```
bryates, line 326: Should we pick a more descriptive naming scheme for the variations, say `nominal`, `up`, `down` instead of `0`, `1`, `-1`?

bryates (changes requested):
I noticed a few changes I did not understand, but maybe these were just from previous merges with the master.

analysis/topEFT/topeft.py
kmohrman: @andrea21z, I am not sure if I understand why we need to pass this argument to the processor? For the other systematics, we either calculate the up/down if the `--do-systs` flag is on, or we skip the up/down if that flag is off. I am wondering if there is a reason why we'd want to handle this systematic in a special way (i.e. be able to turn it on or off separately from the others)?

analysis/topEFT/topeft.py
kmohrman: @bryates, I think these lines were removed because they are no longer necessary (though @andrea21z should of curse correct me if I'm misremembering). Andrea is now passing e.g. `pDataUP/pMC` to the `add()` method [here](https://github.com/TopEFT/topcoffea/blob/adding_MuonScaleCor/analysis/topEFT/topeft.py#L346), instead of calculating this variable beforehand. This was to work around a "feature" of `class coffea.processor.Weights`, where the input values can be modified in place by `add()` (documented [here](https://coffeateam.github.io/coffea/api/coffea.processor.Weights.html?highlight=weight#coffea.processor.Weights.add)).

andrea21z: No, I removed them because we are introducing these calculations directly in the `weights_dict.add` to fix the problem we found some weeks ago.

bryates: Ok, thanks. I guess I didn't follow the exact code that went into that fix.

analysis/topEFT/topeft.py
GonzalezFJR: Well, this uncertainty affects the muon selection, so has to be propagated to the jet selection and all the categories (it's not as simple as changing a weight). The same for the jet energy scale and resolution uncertainties. So one option is re-running the full analysis changing the pT of the muons , as in Andrea's version (there are other analysis with coffea that do this). The other option would be adding tags for the muon pT to basically everything in the analysis (categories, number of jets, HT...) and fill the nominal/up/down axis depending on the chosen muon pT. A third option would be implementing a 'for' loop at the beginning of the analysis, before the muon selection, and remove from the loop those elements that are not affected by muon pT (for example, the electron selection).

kmohrman: Thanks for explaining this @GonzalezFJR. I think it makes sense to merge the PR with the way it's currently implemented. However, I think the third option you described above sounds like what we would want for the long term solution. We could then handle the other similar systematics (jet energy corrections, and maybe others?) as part of the same loop. That way, we could run everything all together in one run of the processor, instead of having to run separately for some of the systematics. I am wondering what your thoughts would be about this? If @GonzalezFJR and @bryates and @andrea21z agree that option 3 is the way we want to handle this, then I'm wondering if this is something @andrea21z you would be interested in working on?

However, I should reiterate that I would be in favor of merging this PR as it currently stands (and updating the handling of those systematics in a later branch). I can make an issue for the discussion of the potential updates to the way these systematics are handled, so that we can move the discussion there, since we'll be merging this PR soon, so it would not be very easy to continue the discussion here.

kmohrman: Here is the issue: #123.

kmohrman (approved):

analysis/topEFT/topeft.py
bryates: I'm in favor of merging now, and following up with the issue later.

andrea21z: Sounds good. I can start to work on the implementation of the loop if you think that is the best option.

PR #122, opened 2021-11-04
Adding sum of weights processor and json file updater

Andrew42:
This PR introduces a new [simple processor](https://github.com/TopEFT/topcoffea/blob/sow-processor/analysis/topEFT/sow_processor.py) that can be run to generate a simple sum of weights histogram that can be used to update the "nSumOfWeights" value in the corresponding json sample file.

Also introduces an [update_jsons.py](https://github.com/TopEFT/topcoffea/blob/sow-processor/analysis/topEFT/update_jsons.py) script, which can be used to process the generated sow histogram file. This script will automatically search for any json files in a specified directory whose name matches a sample name from the histogram file(s). it then overwrites the original json file to make updating as straightforward as possible.

One potential issue with merging this branch is the fact that I pulled in some commits from Ben Tovar's branch where he implemented some fixes to allow installing uproot 4.1.6 via pip, documented in PR #120. While testing out his fixes I had to make some additional bug fixes on top of his commits (specifically 583f0b7) which is likely where the merge conflicts are coming from. As the bug fixes I made are probably very similar to the ones that Ben implemented into his version that actually got merged, I imagine resolving these merge conflicts should be pretty straightforward.


bryates (approved):
Looks good to me. I think we all agreed last week that this will not affect our current setup, so we can merge this now.

topcoffea/json/signal_samples/private_UL/UL16APV_tHq_b1.json
```diff
@@ -59,7 +59,7 @@
     "/store/user/kmohrman/FullProduction/FullR2/UL16APV/Round1/Batch1/naodOnly_step/v1/nAOD_step_tHq4f_all22WCsStartPtCheckdim6TopMay20GST_run0/NAOD-00000_8595.root"
   ],
   "nEvents": 7500000,
-  "nGenEvents": 15024000.0,
```
kmohrman, line 62: @Andrew42, why do we have to get rid of the point 0?

kmohrman:
@Andrew42, have a question about `python_package_run_with_retries`. I'm wondering why it is includes as part of this PR? That was a temporary script that Ben gave us to avoid a bug that has since been corrected upstream. I don't think we need this file anymore, unless I'm missing something.

analysis/topEFT/run_sow.py
```diff
@@ -0,0 +1,184 @@
...
+    'transactions_log': 'tr.log',
+    'stats_log': 'stats.log',
+
+    'environment_file': topeftenv.get_environment(),
+    'extra_input_files': ["sow_processor.py"],
+
+    'schema': NanoAODSchema,
+    'skipbadfiles': False,
+
+    'wrapper': "python_package_run_with_retries",
```
kmohrman, line 107: This should not be needed, I think.

topcoffea/json/signal_samples/private_UL/UL16APV_tHq_b1.json
Andrew42: that's just how the python `json` module is loading and saving the file. I made no requirements on the typing of the dictionary values when I loaded the json. So this is all a result of how the python module is loading and saving the file.

analysis/topEFT/run_sow.py
Andrew42: this is a good point. I'll swap the wrapper back to the original.

topcoffea/json/signal_samples/private_UL/UL16APV_tHq_b1.json
kmohrman: Ok, I don't think we use this number for anything anyway, so probably it's not important. I'm always just a big paranoid about integers because if they get included in a calculation at some point, python might accidental think we want the result to be an integer....

PR #125, opened 2021-11-05
Slimming down the fitting model

sscruz:
Removing WCs per process that don't have an effect on that process. This is done by checking the size of the linear, quadratic and cross-terms. The list of WCs that are kept are stored in a json file (per category) that is used as input when running the text2workspace command.

### DIFF
--- a/analysis/topEFT/datacard_maker.py
+++ b/analysis/topEFT/datacard_maker.py
@@ -8,6 +8,7 @@ import numpy as np
 import os
 import re
 import json
+from copy import deepcopy 
 
 from ROOT import TFile, TH1D, TH2D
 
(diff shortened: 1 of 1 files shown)


analysis/topEFT/datacard_maker.py
```diff
@@ -299,7 +300,7 @@ def getHist(d_hists,name):
             xmax = h.GetXaxis().GetXmax()
             xwidth = h.GetXaxis().GetBinWidth(1)
             h.GetXaxis().SetRangeUser(xmin, xmax + xwidth) #Include overflow bin in ROOT
-            return h
+            return deepcopy(h) # to protect d_hists from modifications
```
bryates, line 303: Interesting. Was this a bug before, or are these improvements the only thing that is sensitive to modifications?

sscruz: I *think* only the new improvements are affected. The only problem should show up when one modifies the histogram that is produced, like we do with the `quad` terms. Since I do it once once to see if the WCs have an effect and then another one to actually store the histos, I think now it's necessary to have this

PR #130, opened 2021-11-11
Adjust topeft processor initilization

Andrew42:
This PR re-organizes the initialization method for the topeft processor to be a bit more compact.

The main changes are the input keyword arguments now simply use the `**kwargs` semantics instead of listing out explicitly every single keyword argument. The other change is to how the processor `self._accumulator` object is constructed. Instead of creating the `processor.dict_accumulator()` all in one call, it is now built up in a more programmatic fashion. The PR also modifies the `run.py` and `work_queue_run.py` scripts to ensure that the keyword arguments are passed into the processor init call correctly.

Finally, I also took the liberty of cleaning up the two run scripts to use a consistent indentation, string formatting, and expanded out some of the more compact 'if' blocks.

As all of these changes are largely aesthetic and should have no impact the actual performance of the code, feel free to reject this PR.

### DIFF
--- a/analysis/topEFT/run.py
+++ b/analysis/topEFT/run.py
@@ -170,7 +170,15 @@ if __name__ == '__main__':
   else:
     print('No Wilson coefficients specified')
  
-  processor_instance = topeft.AnalysisProcessor(samplesdict,wc_lst,hist_lst,do_errors,do_systs,split_lep_flavor,skip_sr,skip_cr)
+processor_instance = topeft.AnalysisProcessor(samplesdict,
+    wc_name_lst = wc_lst,
+    hist_lst = hist_lst,
+    do_errors = do_errors,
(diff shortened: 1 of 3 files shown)


bryates (commented):
Looks good. I'll try to run it just to make sure there are no issues on my end.

bryates:
> Looks good. I'll try to run it just to make sure there are no issues on my end.

Everything ran fine, and these [CR](http://www.crc.nd.edu/~byates2/coffea/adjust_20211201_0924/) plots look good.

Andrew42:
Since this PR isn't really high priority and doesn't have a lot of motivation at the moment, I'm going to go ahead and close this PR for now. The changes are made here aren't super involved or complicated, so if we needed to resurrect this at some point I don't think it would be too much trouble.

Issue #87, opened 2021-08-30
Move to NAOD v9


kmohrman:
Our privately produced NAOD currently corresponds to v8 (as we produced it with `CMSSW_10_6_19_patch2 `, using the NAOD configs generated with the `NanoAODv2` versions of the `cmsDriver` commands listed on [this twiki](https://twiki.cern.ch/twiki/bin/view/CMS/PdmVMcCampaigns)). Once the appropriate `cmsDriver` commands for v9 are included on that twiki, we should remake the NAOD.

rgoldouz:
Hi @kmohrman , It is not needed to wait for the twiki to be updated. You can easily find the commands for generating nanoAOD config files from mcm.
Just choose one of the simulate samples and find the submitted PrepID in MCM, for example
https://cms-pdmv.cern.ch/mcm/requests?prepid=TOP-RunIISummer20UL17NanoAODv9-00001&shown=6442455051

in Action column you will find the setup command,
https://cms-pdmv.cern.ch/mcm/public/restapi/requests/get_setup/TOP-RunIISummer20UL17NanoAODv9-00001

So you have the cmsDriver command.

Please let me know if you have any question.

Thanks,
Reza

kmohrman:
Thanks @rgoldouz. That's a good point, and it is something that's been on my to-do list for quite some time (but I have not gotten to it yet). I will definitely let you know of any questions once I attempt to produce the new configs, thanks!

kmohrman:
Closed with PR #148.

PR #129, opened 2021-11-11
Add code for producing json files of skimmed samples

Andrew42:
This PR introduces a new analysis script called [make_skim_jsons.py](https://github.com/TopEFT/topcoffea/blob/make-skim-jsons/analysis/topEFT/make_skim_jsons.py).

The script works by being given a path to a hadoop directory (or directories) that should contain the results of a lobster skim job. For each lobster WF directory that the script finds, it will attempt to find a matching json sample file. Upon a successful match it will load the json file and replace the `files` list with the list of files from the matching lobster skim directory. The script will then save the json to a new file with '_NDSkim' appended to the file name so as to not clobber the original files. Afterwards the script will printout any skims that it could not find a matching json for, as well as any json files that it could not find a matching skim.

The PR also makes some organizational changes to the scripts introduced in PR #122. Specifically it renames the `update_jsons.py` script to `update_json_sow.py` and also moves some of the functionality out of the script and into the `utils.py` module.

Finally, the PR adds a number of skim jsons for the Run 2 datasets. Skims are available for all three data taking years and a corresponding cfg has also been added that points to these new jsons.

The skim jsons were produced by running the following command from the `topcoffea/analysis/topEFT` directory:

`python make_skim_jsons.py /hadoop/store/user/awightma/skims/NanoAOD_ULv8/v1 --json-dir ../../topcoffea/json/data_samples`

PR #131, opened 2021-11-12
Update json files

bryates:
This PR adds better handling of the extension (`ext`) files. Since these are additional statistics to be added the the main samples, they must be in the main json file. I've added a script to merge them together, and adjust `nEvents`, `nGenEvents`, and `nSumOfWeights`.

I've also updated some of the old `nanoAODv8` samples to `v9` as mentioned in issue #87, but there are still a few that do not exist yet.


bryates:
@Andrew42 I didn't redo any of the PSI samples. Are you still using those, or are you focusing on the skims instead?

Andrew42:
> @Andrew42 I didn't redo any of the PSI samples. Are you still using those, or are you focusing on the skims instead?

I've never used the PSI samples. I'm pretty sure those are only used by @sscruz, since we have trouble accessing those files. However, since this is modifying samples that the skims are based on I will probably want to merge in these changes to PR #129 before that PR gets merged into master.

We might also want to think of some mechanism to keep the 'skim' and 'original' json files synchronized. As it is now the skims have their own copy of the `nSumOfWeights` which is probably a good thing since that value _should_ correspond to w/e files were used to make the skim at the time of creation.

kmohrman:
Thanks @bryates! Just wondering if you could explain a bit more about the change for the `ext` samples. I'm not sure if I understand why we need to merge them together at the JSON level. I think the way we were handling it before, we just specified the same name for the sample axis in the histogram, so all of the events will end up in the same histogram. I'm wondering what is incorrect about that approach we were using before?

bryates:
> Thanks @bryates! Just wondering if you could explain a bit more about the change for the `ext` samples. I'm not sure if I understand why we need to merge them together at the JSON level. I think the way we were handling it before, we just specified the same name for the sample axis in the histogram, so all of the events will end up in the same histogram. I'm wondering what is incorrect about that approach we were using before?

I don't think that's enough, because each event is weighted by `1/sumWeights`, so the `ext` samples will have a different weight than the standard samples. This means the low stats is weighted by a larger weight (`1/ smaller sumWeights`). We should be treating them as one set with the same weights.

kmohrman:
> > Thanks @bryates! Just wondering if you could explain a bit more about the change for the `ext` samples. I'm not sure if I understand why we need to merge them together at the JSON level. I think the way we were handling it before, we just specified the same name for the sample axis in the histogram, so all of the events will end up in the same histogram. I'm wondering what is incorrect about that approach we were using before?
> 
> I don't think that's enough, because each event is weighted by `1/sumWeights`, so the `ext` samples will have a different weight than the standard samples. This means the low stats is weighted by a larger weight (`1/ smaller sumWeights`). We should be treating them as one set with the same weights.

But, each `sum of weights` is specific to each MC sample. If the regular and ext samples are from different MC samples, then I'm not sure if I understand why it makes sense to essentially average their sum of weights (and use that average for all events from both samples), but maybe I'm just thinking about this in the wrong way. Maybe @Andrew42 could also comment?

bryates:
> > > Thanks @bryates! Just wondering if you could explain a bit more about the change for the `ext` samples. I'm not sure if I understand why we need to merge them together at the JSON level. I think the way we were handling it before, we just specified the same name for the sample axis in the histogram, so all of the events will end up in the same histogram. I'm wondering what is incorrect about that approach we were using before?
> > 
> > 
> > I don't think that's enough, because each event is weighted by `1/sumWeights`, so the `ext` samples will have a different weight than the standard samples. This means the low stats is weighted by a larger weight (`1/ smaller sumWeights`). We should be treating them as one set with the same weights.
> 
> But, each `sum of weights` is specific to each MC sample. If the regular and ext samples are from different MC samples, then I'm not sure if I understand why it makes sense to essentially average their sum of weights (and use that average for all events from both samples), but maybe I'm just thinking about this in the wrong way. Maybe @Andrew42 could also comment?

The nominal and ext should be thought of as one giant sample. If we scale the nominal by `xsec/sumNominalWeights` and the ext by `xsec/sumExtWeights` then we'll essentially have twice the predicted yield, and still be sensitive to the fluctuations in the low statistics nominal sample. By combining them into one sample, we smooth things out.

bryates:
The CR plots look good to me.
http://www.crc.nd.edu/~byates2/coffea/update_bkg_20211123_1319/

kmohrman:
The PR looks good to me. And thanks for explaining about the sum of weights for the `ext samples`.

kmohrman (approved):

Andrew42 (approved):
I'm also fine with these changes

kmohrman:
@bryates do you want me to merge this, or were you planning to?

bryates:
> @bryates do you want me to merge this, or were you planning to?

I'll do it now.

PR #133, opened 2021-11-16
Datacard subset fix

bryates:
This PR fixes the issue with the datacard maker using `set_wilson_coeff_from_array` instead of `set_wilson_coefficients`. We can now use the `--POI` command.


kmohrman:
Thanks @bryates. I've been using this branch for the past several days, and have not encountered any issues. I have no objections to merging it at this point.

kmohrman (approved):

PR #136, opened 2021-12-02
Variable binning

bryates:
This PR adds the first, stable, version of variable binning for `ptbl` and `ht`. These give smooth scans (currently with `3l onZ 2b 5j`) excluded. Also, since we haven't processed a pkl file with the new `ext` fix yet, I've been adding
```python
if '_ext' not in proc and proc+'_ext' in self.samples: continue
```
after [L197](https://github.com/TopEFT/topcoffea/blob/8550e747a29e8cd0a9e2ccc9cb6316d8a73ae7d0/analysis/topEFT/datacard_maker.py#L197) in the datacard maker to only process extension samples, if they exist.


kmohrman:
I've checked out this branch and tried running the datacard maker with and without systematics. Everything seemed to run fine, so I'd be happy if you wanted to merge the PR at this point.

kmohrman (approved):

PR #140, opened 2021-12-12
CR plotting updates

kmohrman:
Some minor updates to the CR plotting script:
* Do not include non prompt in the 2los tt category
* Do not include ttJets in the 2lss category
* Rename ttGamma "Conv" to be consistent with TOP-19-001

Unrelatedly, I've also removed the "ext" jsons from the MC background cfg file, since those are no longer specified separately since PR #131.

### DIFF
--- a/analysis/topEFT/make_cr_and_sr_plots.py
+++ b/analysis/topEFT/make_cr_and_sr_plots.py
@@ -483,6 +483,9 @@ def make_all_cr_plots(dict_of_hists,year,unit_norm_bool,save_dir_path):
                 hist_mc_integrated = hist_mc_integrated.remove(["Nonprompt"],"sample")
             if hist_cat == "cr_2lss":
                 hist_mc_integrated = hist_mc_integrated.remove(["Ttbar"],"sample")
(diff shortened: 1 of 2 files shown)


bryates (changes requested):
This mostly looks good, just a few stylistic changes. Can we add back the uncertainty bands on the ratio plots? Also, as @sscruz pointed out, we don't need things like DY and single-top now that we have the data-driven non-prompt backgrounds. I don't think we need to remove these from the `cfg` file, but maybe just make a note to skip them, or have something in the code that skips certain samples if the non-prompt category exists.

kmohrman:
Ok, DY has been removed from the 2lss and 3l CRs. Regarding the error bands, I am not sure what to do about those, (as discussed on the mattermost chat, they disappeared when I began plotting data/mc instead of mc/data).

bryates (approved):
@kmohrman thanks for the updates. I don't think we need to hold up this PR for the ratio bands.

kmohrman:
Thanks @bryates, I've created an issue (#141) so that we won't forget about figuring out the errors on the CR ratio plots.

PR #145, opened 2021-12-17
Fixed overflow bins

bryates:
This PR fixes the issue brought up today about the overflow bins. Since combine does not fit the overflow bins, it was missing the last bin in each of our lepton categories. I've updated the datacard maker to ensure the last bin is an actual bin, and not in the overflow.

### DIFF
--- a/analysis/topEFT/datacard_maker.py
+++ b/analysis/topEFT/datacard_maker.py
@@ -9,7 +9,6 @@ import os
 import re
 import json
 from copy import deepcopy 
-from topcoffea.modules.selection import analysis_bins
 
 from ROOT import TFile, TH1D, TH2D
 
@@ -22,6 +21,12 @@ class DatacardMaker():
         self.syst_special = {'charge_flips': 0.3} # 30% flat uncertainty for charge flips
         self.ignore = ['DYJetsToLL', 'DY10to50', 'DY50', 'ST_antitop_t-channel', 'ST_top_s-channel', 'ST_top_t-channel', 'tbarW', 'TTJets', 'tW', 'WJetsToLNu']
         self.skip = {'nonprompt': '4l'} # E.g. 4l does not include non-prompt background
+        # Dictionary of njet bins
+        self.analysis_bins = {'njets': {'2l': [4,5,6,7,10], # Last bin in topeft.py is 10, this should grab the overflow
+                                        '3l': [2,3,4,5,10],
+                                        '4l': [2,3,4,10] },
+                              'ptbl' : [0, 100, 200, 400, 2000],
+                              'ht'   : [0, 100, 200, 300, 400, 2000] }
         self.fin = infile
         self.tolerance = 1e-5
         self.do_nuisance = do_nuisance
@@ -226,9 +231,9 @@ class DatacardMaker():
             pname = self.rename[p]+'_' if p in self.rename else p
             pname.replace('_4F','').replace('_ext','')
             if 'njet' in variable:
-                h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, analysis_bins[variable][channel[:2]]))
+                h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, self.analysis_bins[variable][channel[:2]]))
             else:
-                h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, analysis_bins[variable]))
+                h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, self.analysis_bins[variable]))
             # Save the SM plot
             h_bases = {syst: h_base.integrate('systematic', syst) for syst in self.syst}
             h_base = h_base.integrate('systematic', 'nominal')
@@ -486,7 +491,7 @@ class DatacardMaker():
             # Find the "S+L+Q", "Q", and "S+Li+Lj+Qi+Qj+2Mij" pieces
             selectedWCsForProc[pname[:-1]]=selectedWCs
             for n,wc in enumerate(selectedWCs):
-                if selfdo_sm: break
+                if self.do_sm: break
 
                 # Get the "S+L+Q" piece
                 name = '_'.join([pname[:-1],'lin',wc])
--- a/analysis/topEFT/datacard_maker.py
+++ b/analysis/topEFT/datacard_maker.py
@@ -248,7 +248,8 @@ class DatacardMaker():
             else:
                 export1d(h_sm, 'data_obs', 'sm', fout)
 
-            if not self.do_sm and (p in self.signal or self.rename[p] in self.signal):
+            isSignal = p in self.signal or self.rename[p] in self.signal
+            if not self.do_sm and isSignal:
                 h_lin = h_bases; h_quad = None; h_mix = None
                 for name,wcpt in self.wcs:
                     # Scale plot to the WCPoint
--- a/analysis/topEFT/datacard_maker.py
+++ b/analysis/topEFT/datacard_maker.py
@@ -225,7 +225,7 @@ class DatacardMaker():
             pname = self.rename[p]+'_' if p in self.rename else p
             pname.replace('_4F','').replace('_ext','')
             if isinstance(self.analysis_bins[variable],dict):
-                lep_bin = channel.split('_')[0].replace('ss','')
+                lep_bin = channel.split('_')[0].split('l')[0] + 'l'
                 h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, self.analysis_bins[variable][lep_bin]))
             else:
                 h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, self.analysis_bins[variable]))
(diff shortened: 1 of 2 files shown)


analysis/topEFT/datacard_maker.py
```diff
@@ -13,19 +13,27 @@
...
         self.syst_terms =['LF', 'JES', 'MURMUF', 'CERR1', 'MUR', 'CERR2', 'PSISR', 'HFSTATS1', 'Q2RF', 'FR_FF', 'HFSTATS2', 'LFSTATS1', 'TRG', 'LFSTATS2', 'MUF', 'PDF', 'HF', 'PU', 'LEPID']
         self.syst_special = {'charge_flips': 0.3} # 30% flat uncertainty for charge flips
         self.ignore = ['DYJetsToLL', 'DY10to50', 'DY50', 'ST_antitop_t-channel', 'ST_top_s-channel', 'ST_top_t-channel', 'tbarW', 'TTJets', 'tW', 'WJetsToLNu']
         self.skip = {'nonprompt': '4l'} # E.g. 4l does not include non-prompt background
+        # Dictionary of njet bins
+        self.analysis_bins = {'njets': {'2l': [4,5,6,7,10], # Last bin in topeft.py is 10, this should grab the overflow
+                                        '3l': [2,3,4,5,10],
+                                        '4l': [2,3,4,10] },
+                              'ptbl' : [0, 100, 200, 400, 2000],
+                              'ht'   : [0, 100, 200, 300, 400, 2000] }
```
Andrew42, line 29: Is this way of specifying the bins sensitive to the actual bin edges of the underlying histogram?

analysis/topEFT/datacard_maker.py
```diff
@@ -215,13 +231,9 @@ def export2d(h):
...
-                if   '2l' in channel: h_base = h_base.rebin('njets', hist.Bin("njets",  "Jet multiplicity ", [4,5,6,7]))
-                elif '3l' in channel: h_base = h_base.rebin('njets', hist.Bin("njets",  "Jet multiplicity ", [2,3,4,5]))
-                elif '4l' in channel: h_base = h_base.rebin('njets', hist.Bin("njets",  "Jet multiplicity ", [2,3,4]))
-            if 'ht' in variable:
-                h_base = h_base.rebin('ht', hist.Bin("ht", "H$_{T}$ (GeV)", [0, 100, 200, 300, 400, 2000]))
-            if 'ptbl' in variable:
-                h_base = h_base.rebin('ptbl', hist.Bin("ptbl", "$p_{T}^{b\mathrm{-}jet+\ell_{min(dR)}}$", [0, 50, 100, 200, 400, 2000]))
+                h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, self.analysis_bins[variable][channel[:2]]))
+            else:
+                h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, self.analysis_bins[variable]))
```
Andrew42, line 236: 1. The `channel[:2]` slicing feels really fragile, but I see that this sort of thing is used elsewhere. Just makes it very reliant on the channel naming following a specific form.
2. I didn't notice earlier when you defined `self.analysis_bins`, but the fact that one set of keys has a `list` as value and the other has a `dict` as the value makes me think that over specializing this dictionary. I'm guessing the reasoning here was b/c `ptbl` and `ht` are meant to have the same binning for all channels, but what if we want to change that? or what we had some other distribution that we want to have different bins for different channels? My suggestion here would be one of two approaches. The first would be to change `self.analysis_bins` to _always_ have the value be a dictionary and if you don't want to have to specify the same exact list for every channel of `ptbl` and `ht`, just add a `default` key that is used if `channel[:2]` can't be found. The other approach would be to instead change the `if 'njet' in variable` check to instead be `if isinstance(self.analysis_bins[variable],dict)`. This way you're basically saying that if we specify a list that means the binning is the same for all channels, otherwise it is channel specific. I'm personally in favor of the 2nd method since it keeps the original dictionary structure as is and simultaneously removes the hard-coded dependence on the `'njet'` string.

analysis/topEFT/datacard_maker.py
```diff
@@ -241,7 +253,7 @@ def export2d(h):
             else:
                 export1d(h_sm, 'data_obs', 'sm', fout)
 
-            if p in self.signal or self.rename[p] in self.signal:
+            if not self.do_sm and (p in self.signal or self.rename[p] in self.signal):
```
Andrew42, line 256: could we define a nicely named variable for `(p in self.signal or self.rename[p] in self.signal)` to make this if statement easier to mentally parse?

analysis/topEFT/datacard_maker.py
```diff
@@ -13,19 +13,27 @@
 from ROOT import TFile, TH1D, TH2D
 
 class DatacardMaker():
-    def __init__(self, infile='', lumiJson='topcoffea/json/lumi.json', do_nuisance=False, wcs=[], single_year=''):
+    def __init__(self, infile='', lumiJson='topcoffea/json/lumi.json', do_nuisance=False, wcs=[], single_year='', do_sm=False):
         self.hists = {}
         self.rename = {'tZq': 'tllq', 'tllq_privateUL17': 'tllq', 'ttZ': 'ttll'} #Used to rename things like ttZ to ttll and ttHnobb to ttH
+        self.rename = {**self.rename, **{'ttH_centralUL17': 'ttH', 'ttH_centralUL16': 'ttH', 'ttH_centralUL18': 'ttH', 'ttHJetToNonbb_M125_centralUL16': 'ttH', 'ttHJetToNonbb_M125_APV_centralUL16': 'ttH', 'ttW_centralUL17': 'ttW', 'ttZ_centralUL17': 'ttZ', 'tZq_centralUL17': 'tllq', 'ttH_centralUL17': 'ttH', 'ttW_centralUL18': 'ttW', 'ttZ_centralUL18': 'ttZ', 'tZq_centralUL18': 'tllq', 'ttH_centralUL18': 'ttH'}}
```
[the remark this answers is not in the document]
bryates: I can't remember now why I originally didn't use `update`, but this method basically does the same thing. I can switch over if people don't like the `**` notation.

analysis/topEFT/datacard_maker.py
bryates: `h.axis('njets').edges()[-1]` seems to do it, I'll switch to that.

analysis/topEFT/datacard_maker.py
bryates: I also like option 2, and I switched to a regex to find the jet bin number.

analysis/topEFT/datacard_maker.py
bryates: Done.

analysis/topEFT/datacard_maker.py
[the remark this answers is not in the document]
Andrew42: It's probably fine to leave as is, I was just surprised to see it and had to stop and think for a second about what was going on, but it's not too hard to figure out if you know what the `**` notation is doing.

kmohrman:
Thanks @bryates for this PR. I've checked it out, but ran into a few issues: For jobs 4, 5, 6, and 7, I got and  error like this in the condor log files:
```
  File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/check_PRs/byates_njets/topcoffea/analysis/topEFT/datacard_maker.py", line 737,
    card.analyzeChannel(**d)
  File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/check_PRs/byates_njets/topcoffea/analysis/topEFT/datacard_maker.py", line 229,
    h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, self.analysis_bins[variable][lep_bin]))
KeyError: '3l1b'
```
I'm wondering if you know what might be wrong? One thing I should mention is that the pkl file I was using to test was a bit old, so it's possible that I just need to run with a newer one; if you've been able to run without getting this issue, then it might be on my end. 

Also, I have two general questions:
1. I'm a bit surprised that the reference file for the CI did not have to change after you fixed the bug for the highest njets bin. I think one of the ref datacards is njets, right? If so, I'm wondering if you could comment on whether or not this is expected. 
2. Sorry if I just missed this, but I'm wondering if you've rerun the fits to obtain updated baseline njets limits?

bryates:
> Thanks @bryates for this PR. I've checked it out, but ran into a few issues: For jobs 4, 5, 6, and 7, I got and error like this in the condor log files:
> 
> ```
> Traceback (most recent call last):
>   File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/check_PRs/byates_njets/topcoffea/analysis/topEFT/datacard_maker.py", line 737,
>     card.analyzeChannel(**d)
>   File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/check_PRs/byates_njets/topcoffea/analysis/topEFT/datacard_maker.py", line 229,
>     h_base = h_base.rebin(variable, hist.Bin(variable,  h.axis(variable).label, self.analysis_bins[variable][lep_bin]))
> KeyError: '3l1b'
> ```
> 
> I'm wondering if you know what might be wrong? One thing I should mention is that the pkl file I was using to test was a bit old, so it's possible that I just need to run with a newer one; if you've been able to run without getting this issue, then it might be on my end.

Thanks for catching that. I've fixing the issue by changing how I get the lepton category. 
> 
> Also, I have two general questions:
> 
> 1. I'm a bit surprised that the reference file for the CI did not have to change after you fixed the bug for the highest njets bin. I think one of the ref datacards is njets, right? If so, I'm wondering if you could comment on whether or not this is expected.

I think this is because I use `Integral()` which _does_ use the overflow. The only difference was combine was ignoring that last bin before. 
> 2. Sorry if I just missed this, but I'm wondering if you've rerun the fits to obtain updated baseline njets limits?

No, but I should! I'll get that started now.

kmohrman:
Ok, I've rerun on a test pkl file after pulling your changes, and it ran without any errors. I'd be happy if you wanted to merge it at this point.

Also, just for future reference (and so the comparison will be linked with this PR), I've created an Issue (#147) to follow up on the comparison of the baseline njets fit before and after the bug fix.

PR #151, opened 2022-01-24
Taking sum of weights at SM from json

sscruz:
This PR forces all the modules to use the sum of weights at SM for EFT samples from the value encoded in the json file, instead of computing the sum of weights at runtime. This allows to use skims on EFT samples transparently


sscruz:
Please wait before checking this PR, something odd happened when rebasing with respect to `master`

sscruz:
I think it's fixed now, I had to upload the sum of weights in the .json file that is used as an input for the CI checks

bryates (commented):
Looks good. I see the CI is failing at the yields check. Can you update the reference json files?

kmohrman:
Thanks for the PR @sscruz. I haven't had a chance to look through it in detail yet, but I just wanted to comment on what @bryates mentioned:
> I see the CI is failing at the yields check. Can you update the reference json files?

I don't think the changes in this PR (at least in the way I understand it) should affect the yields, it should just affect the way we calculate the yields. So I think we might want to figure out why the yields are changing rather than updating the reference.

bryates:
> Thanks for the PR @sscruz. I haven't had a chance to look through it in detail yet, but I just wanted to comment on what @bryates mentioned:
> 
> > I see the CI is failing at the yields check. Can you update the reference json files?
> 
> I don't think the changes in this PR (at least in the way I understand it) should affect the yields, it should just affect the way we calculate the yields. So I think we might want to figure out why the yields are changing rather than updating the reference.

I'm realizing now it might not be the yields, but just the json file itself. It was a file comparison that failed.

topcoffea/json/test_samples/UL17_private_ttH_for_CI.json
```diff
@@ -33,6 +33,6 @@
   ],
   "nEvents": 7993714,
   "nGenEvents": 40553247,
-  "nSumOfWeights": 40553247.0,
+  "nSumOfWeights": 575.37356,
```
kmohrman, line 36: Just a quick question on this, did this new number come from summing the weights for all of the events in this file? 

I guess the original numbers here are wrong because for this ref JSON I think we just copied one of the existing JSONs and swapped in the ROOT file we were interested in (so the `nEvents`, `nGenWeights`, and `nSumOfWeights` would still correspond to whatever was calculated for the original set of samples included in the JSON we were copying). If so, that'd mean the `nSumOfWeights` is now correct, but the other numbers are still wrong? However, maybe we don't really care so much about that since I don't think we use those numbers for anything. But even so, maybe we should consider replacing those with numbers that are clearly just placeholders (e.g. 0 or -1) or just grab the right numbers and put them in. Though I think this is probably not very important.

kmohrman:
Thanks @sscruz for this PR! These changes look good to me. I've checked out the branch and tried a test job, and it seemed to work without any issues. Apart from my one question above (which probably does not really require any changes anyway), I have no objections to merging this branch.

topcoffea/json/test_samples/UL17_private_ttH_for_CI.json
sscruz: yes, the new number comes from summing the weight of all events in the file. About the other numbers, now that you mention it, I think I don't understand them: what is the difference between `nEvents` and `nGenEvents`? This file is not skimmed, is it? So then they should be the same

kmohrman: Hmm, those numbers should come from what's stored in the NAOD file, which we get using the [GetAllInfoFromFile](https://github.com/TopEFT/topcoffea/blob/master/topcoffea/modules/fileReader.py#L196) function of `topcoffea.modules.fileReader` (though I think those numbers would correspond to the sum of all of the files that were originally in the JSON we copied, not just this one file). Anyway, assuming the right numbers are being stored in the NAOD, I guess the reason why nGen and nEvents are different could be due to the fact that this is a ttH+j sample? (Since the matching procedure throws out a significant fraction of the LHE events during the GEN step to remove the overlap between the ME and PS, so maybe the nGenEvents number is from before the matching?)

sscruz: hmm, I'm not sure: looking at the code it seems `nEvents` is the entries in the `Events` `TTree`, where `nGenEvents` is the sum of all the entries of the `genEventCount` branch. These two numbers should agree in an unskimmed sample and in fact they do for this specific file.
Anyway, as you said, these two numbers are not used anywhere in the analysis...

kmohrman: Hmm, yeah, I am not sure. It seems in general these numbers do not agree for our private samples [1], while for the central samples they do [2]. It's odd, but again not actually used anywhere... maybe we should make an Issue so we can look into it if we have time at some point (or if we ever end up wanting to use those numbers) and just move on for now?

[1] https://github.com/TopEFT/topcoffea/tree/master/topcoffea/json/signal_samples/private_UL
[2] https://github.com/TopEFT/topcoffea/tree/master/topcoffea/json/background_samples/central_UL

kmohrman:
As discussed above, we should at some point understand `nEvents` and `nGenEvents` listed in the JSON files for the private samples. However, these numbers are not currently being used for anything, so it should no cause a problem, and should not hold up this PR (also, the problem was not introduced by this PR, it was just noticed in the review of this PR). We've created an issue (#154) to remind ourselves of the potential problem.

With that out of the way, I'd be happy to merge the PR now as long as no has any objections.

PR #152, opened 2022-01-25
Allow for different categories for different njet bins

kmohrman:
This PR rearranges how we define the selection categories in the processor (it is not supposed to change any of the outputs, just the way the code is internally structured). Previously, the `sr_cat_dict` (and `cr_cat_dict`) had been structured in such a way so as to require that all of the channels in the `lep_chan_lst` would all have the same njet bins. This PR reorganizes the `sr_cat_dict` (and `cr_cat_dict`) so that each njet bin may have a different list of selection channels.

Motivation for this PR:
*  Many of the optimization strategies we have been discussing involve different categories for different njet bins. This reorganization should make that easier to implement. 
* I had already implemented these changes in the branch I was working on for the hadronic top studies. However, as it seems that branch will not end up getting merged (since the hadronic top pt does not actually seem to improve our sensitivity), I thought that it might be good to get these changes merged since they are really independent of the hadronic top studies, are already written, and might be useful in the future.

### DIFF
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -496,50 +496,110 @@ class AnalysisProcessor(processor.ProcessorABC):
           # This dictionary keeps track of which selections go with which SR categories
           sr_cat_dict = {
             "2l" : {
-                "lep_chan_lst" : ["2lss_p" , "2lss_m"],
-                "lep_flav_lst" : ["ee" , "em" , "mm"],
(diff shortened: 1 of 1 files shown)


bryates (approved):
Looks good, thanks for making this PR!

kmohrman:
Thanks @bryates, I'll go ahead and merge it (hopefully will not cause any conflicts for any of the other open PRs).

PR #146, opened 2021-12-20
Update skims for data samples using NAOD v9

Andrew42:
This PR updates and makes available skims for the Run 2 datasets using NAODv9. It follows the same procedure and setup as in PR #129. It updates the json making script to produce the json files for the NAODv9 data and replaces the old data jsons with the new ones. These jsons were then used to produce the skimmed datasets using the [LobsterSkimming](https://github.com/TopEFT/LobsterSkimming) code.

Once the skims were produced, jsons for these samples were made by running the following command from the `topcoffea/analysis/topEFT` directory:
```
python make_skim_jsons.py --json-dir ../../topcoffea/json/data_samples --file ND_data_skim_locations.txt
```
Which uses an updated version of the `ND_data_skim_locations.txt` file, also included in this PR.

The new jsons (skimmed and un-skimmed) fully replace the previous ones and any old files have been completely removed. This is due to fact that naming for a few of the 2016 run eras has changed w.r.t. the NAODv8 versions. The changes are the following:
```
*_C-UL2016 -> *_C-HIPM_UL2016
*_D-UL2016 -> *_D-HIPM_UL2016
*_E-UL2016 -> *_E-HIPM_UL2016
```
By fully removing the old data jsons we ensure that there aren't any stray json files pointing to old NAODv8 data lying around. In any case, the `data_samples_NDSkim.cfg` and `data_samples.cfg` files have both been updated to reflect the new naming and should point to the correct json file locations.

PR #158, opened 2022-01-31
Add json files for 2016 MuonEG dataset

Andrew42:
This PR adds the json files for the MuonEG dataset that were missing from the 2016 data samples.

I've only included the generated json files in this PR, without updating `make_jsons.py` and also without updating the corresponding `cfg` files. This is because this dataset corresponds to the NAOD-v9 version, which is being handled more generally in PR #146, which already includes the naming changes for those files (but will still need to be updated to include this missing dataset).

I think it will be easier to include just the missing MuonEG json files here and then have these changes merged back into the #146 afterwards. This should hopefully minimize any confusing merge conflicts that might show up between the two PRs.

As already mentioned, this PR introduces the dataset files corresponding to _NAOD-v9_. It should be noted that the current NAOD data samples on `master` are for NAOD-v8 (at least until PR #146 is merged). This means that we probably shouldn't actually include this dataset in any CR studies/plots until after #146 is finally merged.


kmohrman:
This looks like a straightforward PR to merge. I agree with your proposal for how to proceed (i.e. merging this PR, but not using it till #146 is merged). I'd be happy to merge this PR unless anyone else has any objections.

bryates (approved):
Looks good to me. We can merge now and not use it until #146 is merged, or just fold this into #146. I'm fine with whatever is easier.

PR #159, opened 2022-01-31
CI update on separating pytest runs 

ywan2:
Changing `conda run -n coffea-env pytest --cov=./ --cov-report=xml -rP` to `conda run -n coffea-env pytest tests/[filename].py` to run a specific test file in a step. All test files are now running individually in separate steps under `Full_Analysis`.

Creating a new test file called `test_workqueue.py` to single out the run of WQ executor in test file `test_topcoffea.py` , and changing the original `test_topcoffea.py` to `test_futures.py` to only run futures executor. Each is included in a single step under `Full_Analysis`.


.github/workflows/main.yml
```diff
@@ -34,11 +34,35 @@ jobs:
         run: |
           wget --no-verbose http://www.crc.nd.edu/~kmohrman/files/root_files/for_ci/ttHJet_UL17_R1B14_NAOD-00000_10194.root
 
-      - name: Run processors over test files with futures executor
+      - name: Test unit
         run: |
           conda install -y -n coffea-env -c conda-forge pytest pytest-cov root_base
           mkdir histos
-          conda run -n coffea-env pytest --cov=./ --cov-report=xml -rP
```
bryates, line 41: This change removes our code coverage check. If you add back the `--cov=./ --cov-report=xml -rP` part to the _first_ `pytest` line, and `--cov=./ --cov-report=xml -rP --cov-append` should hopefully work.

bryates:
Thanks for starting this @ywan2, but the CI doesn't run on a private fork. Could you push this to a `pytest_check` branch on the TopEFT repo instead?

kmohrman:
@bryates, all of the CI tests actually do run on @ywan2's fork (except for the "Upload codecov" one), so for me, I'm fine with it being on her fork. I think in the past we have not worried about that "Upload" step when reviewing PRs from forks (especially since we don't really even use the codecov tests at this point). Is there a reason why you suspect it might be important to check the "Upload" step in this case?

bryates:
> @bryates, all of the CI tests actually do run on @ywan2's fork (except for the "Upload codecov" one), so for me, I'm fine with it being on her fork. I think in the past we have not worried about that "Upload" step when reviewing PRs from forks (especially since we don't really even use the codecov tests at this point). Is there a reason why you suspect it might be important to check the "Upload" step in this case?

Ok, good point. As for further clean-up, I recently found the `--quiet` command for `conda install`. This should remove all the extraneous installation printouts on the stage that installs `ROOT`.

ywan2:
I updated two things:
1) I separated `conda install ...root_base` from `test unit`step and have it included in `Pytest setup` step (a new step I created). 
2) I added line `--cov=./ --cov-report=xml -rP` within the first pytest step (test unit).
However, it still doesn't fix the issue of Codecov.

What do you think? Thank you for all the great comments!

bryates:
> I updated two things:
> 
> 1. I separated `conda install ...root_base` from `test unit`step and have it included in `Pytest setup` step (a new step I created).
> 2. I added line `--cov=./ --cov-report=xml -rP` within the first pytest step (test unit).
>    However, it still doesn't fix the issue of Codecov.
> 
> What do you think? Thank you for all the great comments!

I think we need to add `--cov=./ --cov-report=xml -rP --cov-append` to all the other `pytest` commands, and that should allow them to write to the same report. It still won't upload to Codecov since this is on a fork, but that's not a big deal. Once we merge this PR, it will run and hopefully upload correctly. 

If you want to test it locally, you can try running all the `pytest` commands with `html` instead of `xml`. That will generate a web folder you can put in your `~/www` directory on the ND T3 to browse.

.github/workflows/main.yml
```diff
@@ -34,11 +34,38 @@ jobs:
...
+        run: |
+          conda run -n coffea-env pytest tests/test_make_1d_quad_plots.py 
+
+      - name: Run processors over test files with futures executor
+        run: |
+          conda run -n coffea-env pytest tests/test_futures.py
+
+      - name: Run processors over test files with workqueue executor
+        run: |
+          conda run -n coffea-env pytest tests/test_workqueue.py
```
bryates, line 68: I think once we append `--cov=./ --cov-report=xml -rP --cov-append` to these test, this PR should be good to go.

.github/workflows/main.yml
```diff
@@ -50,6 +77,11 @@ jobs:
           cd analysis/topEFT
           conda run -n coffea-env python comp_yields.py output_check_yields.json test/UL17_private_ttH_for_CI_yields.json -t1 "New yields" -t2 "Ref yields"
 
+      - name: Test datacard maker
+        run: |
+          conda run -n coffea-env pytest tests/test_datacardmaker.py
```
bryates, line 82: `--cov=./ --cov-report=xml -rP --cov-append` here too.

bryates (changes requested):

.github/workflows/main.yml
bryates, line 82: Sorry, one more place we need `--cov=./ --cov-report=xml -rP --cov-append`

bryates (approved):
Thanks @ywan2 for implementing all the changes! I know there were a lot of small tedious ones. At this point, I'm happy with merging this PR.

ywan2:
> Thanks @ywan2 for implementing all the changes! I know there were a lot of small tedious ones. At this point, I'm happy with merging this PR.

Thank you @bryates  for noticing the small details and checking the code!

kmohrman:
Thanks @ywan2 for the PR, and @bryates for the thorough review! It all looks good to me at this point, so I'd be happy for it to be merged at this point.

PR #160, opened 2022-01-31
Only loop over the systs that we need to

kmohrman:
This PR modifies the list of systematics that we loop over in the "outside" loop in the processor to limit this list to only those that should impact the selection. This should significantly improve the runtime when running the processor with the `_do_systematics ` option turned on (since we won't have to rerun all of the code in the processor for _all_ of the systematics). 

As a mostly unrelated update, the behavior of the `ApplyJetSystematics()` function of `corrections` has also been updated so that the `else` will now raise an exception. This is to avoid the scenario where we e.g. pass something like `JERup` to the function (when we had meant to pass `JERUp`). Currently this would result in the code silently returning `cleanedJets` instead of `cleanedJets.JER.up`. With this update, the code would now complain about an unknown `syst_var`.

### DIFF
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -172,7 +172,7 @@ class AnalysisProcessor(processor.ProcessorABC):
         # Update muon kinematics with Rochester corrections
         mu["pt_raw"]=mu.pt
         met_raw=met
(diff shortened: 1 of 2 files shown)


kmohrman:
Thanks @andrea21z! @bryates, should we go ahead and merge it now, or do you have any comments/objections?

PR #164, opened 2022-02-01
wq install: use --strict-channel-priority

btovar:
It gives a hint to conda that makes environment creation and package
installation a little bit faster.

This will be the default in an upcoming conda version.


kmohrman:
Looks good to me!

PR #166, opened 2022-02-03
Adding new leading object pt hists

kmohrman:
This PR adds several new histograms with the variables we have been looking into throughout the ongoing optimization studies. The PR is not very invasive, as it mainly just adds new histograms (and e.g. does not change anything about the selection or categories).

The PR also includes a few minor modifications to the datacard maker to add an option that lets us specify which variables to make cards for. By default, the datacard maker will make cards for all variables in the pkl file which are included in a "known_var_lst" list in the datacard maker (which should correspond to the variables for which we've defined an analysis binning).


tests/test_futures.py
```diff
@@ -80,6 +84,10 @@ def test_datacard_2l_ht():
         "python",
         "analysis/topEFT/datacard_maker.py",
         "analysis/topEFT/histos/output_check_yields.pkl.gz",
+        "--var-lst",
+        "njets",
+        "ht",
+        "ptbl",
         "-j",
         "9"
```
bryates, line 92: It wont make a difference, but we could also just use e.g. `HT` here (since it is the check `HT` function) and change 9 to 0 (since 9 was originally `HT 2lss_p`. Feel free to keep it how it is now if figuring out the `-j` numbers is too much of a headache.

analysis/topEFT/datacard_maker.py
```diff
@@ -706,17 +730,34 @@ def condor_job(self, pklfile, njobs, wcs, do_nuisance, do_sm):
...
+    # If we don't specify a variable, use all variables in the pkl that we have a binning defined for
+    include_var_lst = []
+    target_var_lst = var_lst
+    if len(var_lst) == 0:
+        target_var_lst = yt.get_hist_list(pklfile)
+    for var_name in target_var_lst:
+        if var_name in card.known_var_lst:
+            include_var_lst.append(var_name)
+    if len(include_var_lst) == 0: raise Exception("No variables specified")
+    print(f"\nMaking cards for: {include_var_lst}\n")
```
bryates, line 756: Awesome! I've been wanting to do this for a while.

bryates (approved):
Looks good. Just some _minor_ comments, which you can feel free to ignore. I'll go ahead and approve this.

tests/test_futures.py
kmohrman: Thanks, if it's ok with you I'd probably prefer to just leave it as it is so we don't have to figure out the `-j` numbers.

bryates: Sounds good.

PR #169, opened 2022-02-04
allow http or https as servers for root files

btovar:


bryates (approved):
Looks goo to me. If @kmohrman or no one else has any objections, we can merge this.

kmohrman:
Looks good to me.

PR #170, opened 2022-02-05
Datacard maker overflow updates

kmohrman:
This PR fixes the overflow issue in the datacard maker. The issue was that the overflow bins in the coffea histograms were being included as overflow in the ROOT histograms as well. This is problematic since combine ignores overflow bins.

There are probably many ways of handling this issue, but the approach taken in this PR is to perform the modifications on the ROOT histograms (as opposed to the coffea histograms) at the start of the `makeCardLevel()` method. For each hist, the contents of the overflow bin are added to the last bin in the hist, the contents of the overflow bin's error is added in quadrature to the last bin's error, the overflow bin's contents are set to 0, and the overflow's error is set to 0.

### DIFF
--- a/analysis/topEFT/datacard_maker.py
+++ b/analysis/topEFT/datacard_maker.py
@@ -334,7 +334,17 @@ class DatacardMaker():
             xmin = h.GetXaxis().GetXmin()
             xmax = h.GetXaxis().GetXmax()
             xwidth = h.GetXaxis().GetBinWidth(1)
-            h.GetXaxis().SetRangeUser(xmin, xmax + 1.5*xwidth) #Include overflow bin in ROOT
+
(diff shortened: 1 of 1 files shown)


kmohrman:
Just wanted to follow up on this and mention that I reran some fits with this update, and the impact on the limits was not large. I checked a few different variables (`njets`, `ptbl`, `lj0pt`,`lj0ptptz`) and the differences from before and after the overflow fix were quite small, the largest difference I saw was about 2.5% (for cQq83 for the `lj0ptptz` fit).

bryates (approved):
Thanks @kmohrman, I think this PR looks good. Once we merge it, I'll modify #167 to _just_ address the `syst_special` part.

kmohrman:
Sounds good, thanks! Should I go ahead and merge it now?

PR #171, opened 2022-02-07
Trigger SFs, new UL Lepton and b-tagging SFs

andrea21z:
This PR includes:

- Trigger SFs: by channel and by year. For 2l as a function of l0pt and l1pt, and for 3l as a function of l0pt and l0eta with less bins. Only stats. uncertainty is included.
- UL lepton SFs (done by Barbara): muon SFs are final but electron SFs have to be updated when new electron MVA is included in the code.
- New UL b-tagging SFs, also for 2016 and 2016APV.
- Fixing some bugs.


andrea21z:
@bryates do you understand why it is failing?

bryates:
> @bryates do you understand why it is failing?

I do not. It has to do with the workqueue test, which is strange. I ran the CI a couple of times to make sure it wasn't a fluke. I'll try looking into it more.

kmohrman:
Have not looked at it, but if it's the WQ test that's failing I'm guessing that there have been new data files (e.g. txt or root) added to the `topcoffea/topcoffea` directory that have not been listed in the `setup.py` (as we've discussed previously, these files are not included in the package by default since they're not .py files, so they need to be listed in the setup script. The reason why this is never caught when running locally is because `topcoffea` just points to the local copies when you've installed with `-e`, so it can run fine even though those files are not actually being included properly. To catch these types of omissions ahead of time you can uninstalling and reinstalling topcoffea without the -e, and it should fail in the same way. However, now that we have the WQ CI test, we can also just notice the problems when it fails the CI as well).

bryates:
> Have not looked at it, but if it's the WQ test that's failing I'm guessing that there have been new data files (e.g. txt or root) added to the `topcoffea/topcoffea` directory that have not been listed in the `setup.py` (as we've discussed previously, these files are not included in the package by default since they're not .py files, so they need to be listed in the setup script. The reason why this is never caught when running locally is because `topcoffea` just points to the local copies when you've installed with `-e`, so it can run fine even though those files are not actually being included properly. To catch these types of omissions ahead of time you can uninstalling and reinstalling topcoffea without the -e, and it should fail in the same way. However, now that we have the WQ CI test, we can also just notice the problems when it fails the CI as well).

Good catch! I'm trying to add those now.

bryates:
> > Have not looked at it, but if it's the WQ test that's failing I'm guessing that there have been new data files (e.g. txt or root) added to the `topcoffea/topcoffea` directory that have not been listed in the `setup.py` (as we've discussed previously, these files are not included in the package by default since they're not .py files, so they need to be listed in the setup script. The reason why this is never caught when running locally is because `topcoffea` just points to the local copies when you've installed with `-e`, so it can run fine even though those files are not actually being included properly. To catch these types of omissions ahead of time you can uninstalling and reinstalling topcoffea without the -e, and it should fail in the same way. However, now that we have the WQ CI test, we can also just notice the problems when it fails the CI as well).
> 
> Good catch! I'm trying to add those now.

It seems to be complaining about a file that exists
https://github.com/TopEFT/topcoffea/runs/5110351638?check_suite_focus=true#step:15:158
`topcoffea/data/JEC/Summer19UL17_V5_MC_L1FastJet_AK4PFchs.txt`
and `data/JEC/*.txt` _is_ included in `setup.py`. I'll keep digging.

bryates:
> > > Have not looked at it, but if it's the WQ test that's failing I'm guessing that there have been new data files (e.g. txt or root) added to the `topcoffea/topcoffea` directory that have not been listed in the `setup.py` (as we've discussed previously, these files are not included in the package by default since they're not .py files, so they need to be listed in the setup script. The reason why this is never caught when running locally is because `topcoffea` just points to the local copies when you've installed with `-e`, so it can run fine even though those files are not actually being included properly. To catch these types of omissions ahead of time you can uninstalling and reinstalling topcoffea without the -e, and it should fail in the same way. However, now that we have the WQ CI test, we can also just notice the problems when it fails the CI as well).
> > 
> > 
> > Good catch! I'm trying to add those now.
> 
> It seems to be complaining about a file that exists https://github.com/TopEFT/topcoffea/runs/5110351638?check_suite_focus=true#step:15:158 `topcoffea/data/JEC/Summer19UL17_V5_MC_L1FastJet_AK4PFchs.txt` and `data/JEC/*.txt` _is_ included in `setup.py`. I'll keep digging.

It seems to run ok on my end, even after uninstalling topcoffea and installing _without_ the `-e` flag.

kmohrman:
Thanks @andrea21z for this PR. It looks good to me. I've just updated the ref yields (the ones we compare manually, and are a larger-scale comparison than what the CI does). @andrea21z and @bryates, are there any remaining open items to discuss, or would we be ready to merge once the CI finishes running?

bryates:
> Thanks @andrea21z for this PR. It looks good to me. I've just updated the ref yields (the ones we compare manually, and are a larger-scale comparison than what the CI does). @andrea21z and @bryates, are there any remaining open items to discuss, or would we be ready to merge once the CI finishes running?

I'm fine with merging.

PR #173, opened 2022-02-08
Add path entry to json sample files

Andrew42:
This PR introduces a new entry to the json making code called `path`. This entry is intended to indicate the location from which the root files were originally found.

This is mostly only relevant for the central samples and the unskimmed data, as those files are obtained via their DAS dataset names. The private samples would have their Hadoop '/store` path stored instead, which of course could already be inferred from the list of files in the json record to begin with.

The PR only updates the json files for the data samples. Since this is a pure addition to the repo, _not_ updating the other json files shouldn't pose any issues since nothing should depend on it yet and we can just make piecemeal updates as they become necessary or just naturally due to other future updates to the files.


bryates (approved):
Looks good to me.

kmohrman:
Looks good to me as well.

PR #174, opened 2022-02-09
Add optional Ecut to processor

kmohrman:
This PR adds an option `--ecut` to the run scripts. This option allows us to pass a value to the processor to use as a cut on the ST of an event (events with an ST larger than this value are thrown out). The default value for the cut is `None`, which means the cut will not be applied at all. As discussed at the [Feb 4 2022 weekly group meeting](https://indico.cern.ch/event/1103039/contributions/4723094/attachments/2385185/4076334/preliminary_validity_studies_Feb4_2022.pdf), we can use this option to check the validity of our limits. The initial checks (with cuts of 1 and 2 TeV) indicated that our limits seems to be within the approximate perturbatively limits discussed in arxiv 1802.07237.

Somewhat unrelatedly, this PR also adjusts some of the binning of a few of the hists in the processor (we had been binning *really* finely, so I've made it a bit coarser). Finally, this PR also removes a repeated line from the datacard maker.


analysis/topEFT/topeft.py
```diff
@@ -60,21 +60,23 @@ def __init__(self, samples, wc_names_lst=[], hist_lst=None, do_errors=False, do_
...
-        "nbtagsl" : HistEFT("Events", wc_names_lst, hist.Cat("sample", "sample"), hist.Cat("channel", "channel"), hist.Cat("systematic", "Systematic Uncertainty"),hist.Cat("appl", "AR/SR"), hist.Bin("nbtagsl",  "Loose btag multiplicity ", 5, 0, 5)),
-        "l0pt"    : HistEFT("Events", wc_names_lst, hist.Cat("sample", "sample"), hist.Cat("channel", "channel"), hist.Cat("systematic", "Systematic Uncertainty"),hist.Cat("appl", "AR/SR"), hist.Bin("l0pt",    "Leading lep $p_{T}$ (GeV)", 50, 0, 500)),
+        "nbtagsl" : HistEFT("Events", wc_names_lst, hist.Cat("sample", "sample"), hist.Cat("channel", "channel"), hist.Cat("systematic", "Systematic Uncertainty"),hist.Cat("appl", "AR/SR"), hist.Bin("nbtagsl", "Loose btag multiplicity ", 5, 0, 5)),
+        "l0pt"    : HistEFT("Events", wc_names_lst, hist.Cat("sample", "sample"), hist.Cat("channel", "channel"), hist.Cat("systematic", "Systematic Uncertainty"),hist.Cat("appl", "AR/SR"), hist.Bin("l0pt",    "Leading lep $p_{T}$ (GeV)", 20, 0, 500)),
```
bryates, line 69: Odd binning always make me uncomfortable. If we keep the number of bins at 50, then we have nice 10 GeV wide bins. I suppose it really just depends on how we rebin in the datacard maker.

kmohrman: Sure, I could up it to 50 if you'd prefer that. Though since we end up using just 4 bins at the analysis level, for me wider bins (e.g. 25GeV) make it a bit easier to visualize where to put the bin boundaries for the 4 bins. Also, I was trying to be consistent with the others that have ranges of 0 to 1000 and 40 bins (i.e. also bin widths of 25). I guess 25GeV seemed like a nice number to me (since I wanted something larger than 10, but 50 seemed too coarse). Are you ok with the other histograms with ranges of 0 to 1000 and 40 bins, or would you prefer a different binning for those as well?

kmohrman: Ah, ok, I see from your later comments that you also have the same comment for the other histograms. Anyway, I'd prefer a bin size larger than 10GeV (since as we've seen in the past that can be a bit noisy, as well as make it harder to figure out good analysis level bins), but smaller than 50GeV. I was thinking that 25 was the nicest number between 10 and 50, but maybe you'd have a different suggestions?

bryates: Ok, we can stick with 25 GeV.

bryates (approved):

PR #175, opened 2022-02-10
Fix typos in key names

kmohrman:
This PR fixes some typos in UL16 key names in the `objects.py` file that were noticed when testing with 2016 samples with the systematics turned on.


andrea21z:
> This PR fixes some typos in UL16 key names in the `objects.py` file that were noticed when testing with 2016 samples with the systematics turned on.

Sorry, I forgot to change the btagging in `objects.py`. Now we have UL16 and UL16APV.

andrea21z (approved):

Issue #172, opened 2022-02-07
Recheck HT/ST type variables


kmohrman:
Now that we have fixed an issue with the way the overflow was being handled (PR #170), it would be good to recheck some of the older variables we tried (in case the poor sensitivity we observed with these variables was being affected by the fact that the overflow bins were being ignored). After fixing the overflow issue, we checked the size of the effect for some of the variables we're currently considering (e.g. `lj0pt`) but not the older ones we had previously ruled out. So at some point, we'll probably want to run a quick check over these old variables (that make use of info about the whole event e.g. ST) just to make sure the conclusions do not change (i.e. make sure they still show worse sensitivity than the variables that target an object or pair of objects with high pt).

kmohrman:
Initial checks show that HT still performs worse than our current frontrunners (`lj0pt` and similar). Similarly, the invariant mass variable for leptons and jets (essentially our attempt to get close to something like sqrt(s)) performs worse than `lj0pt`. However, ST actually seems comparable to `lj0pt`. I will post a followup with more details later.

kmohrman:
For future reference, here are the details of this study:
* For these comparisons, I used the Full Run 2 signal samples (no backgrounds, no systematics).
* For these comparisons, I checked three variables that include information about the whole event that we have checked in the past (`ht`, `ljptsum`, and `ljinvmass`) against `lj0pt`, where these variables are defined as follows:
    * `ht`: Sum of pt of all of the jets in the event
    * `ljptsum`: Sum of pt of all of the jets and leptons in the event (aka `st`)
    * `ljinvmass`: Invariant mass of sum of the four vectors of all of the leptons and jets in the event (i.e. our attempt to get at something close to `sqrt(s)`)
    * `lj0pt`: pt of the pair from the leptons+jets collection that has the largest pt (this variable is one that has performed the best in most of our tests so far, so seemed like a good benchmark to compare the `ht` type variables against). 
* What we've seen in the past: All of these variables that include information about all of the particles in the event have generally performed worse than `ptbl` (the benchmark variables we used before finding better variables e.g. `lj0pt`). However, the concern is that this degraded performance may have been caused in part by the overflow issue (as it may have impacted these `ht` type variables more strongly than the `ptbl` type variables because of the binning we were using, i.e. more of the tail of the distribution may have been in overflow for the `ht` type variables).  
* Results:
    * `ht` generally performs worse than `lj0pt` by up to about 45%
[...]

PR #176, opened 2022-02-13
Add S_T hist binning info to datacard maker

kmohrman:
This PR just adds the binning info for the S_T histogram to the datacard maker. As discussed in Issue #172, this variable was found to provide comparable sensitivity to the other variables we've been considering, so it'd be good to have the datacard maker set up to easily produce data cards for this variable. (Note that the variable is called `ljptsum` in the code, since `st` is too hard to search.) 

Unrelatedly, this PR also adjusts the H_T binning to better match the shape of the H_T histogram (and to have 4 bins as the other histograms do). However, since H_T does not provide very good sensitivity, it will probably not be used anyway.

### DIFF
--- a/analysis/topEFT/datacard_maker.py
+++ b/analysis/topEFT/datacard_maker.py
@@ -34,7 +34,7 @@ class DatacardMaker():
         self.year = single_year
         self.do_sm = do_sm
         # Variables we have defined a binning for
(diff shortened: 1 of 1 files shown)


bryates:
Looks good, I'd say it's safe to merge now.

PR #178, opened 2022-02-14
Fix sum of weights for multiple batches

bryates:
This PR merges all the batches (for private samples) into the corresponding `b1` files, e.g.
```
topcoffea/json/signal_samples/private_UL/UL18_ttHJet_b1.json
topcoffea/json/signal_samples/private_UL/UL18_ttHJet_b2.json
topcoffea/json/signal_samples/private_UL/UL18_ttHJet_b3.json
```
are merge into `topcoffea/json/signal_samples/private_UL/UL18_ttHJet_b1.json`, and `nEvents`, `nGenEvents`, and `nSumOfWeights` are summed together.


kmohrman:
Thanks for this PR @bryates. I agree it'd be one way to fix the issue with the sum of the weights. However, I'm not sure if we want to do it in this way. Sometime is it useful to be able to easily process a smaller set of events. I think we should consider potentially adding something to the run script that checks if multiple JSONs correspond to the same sample, and just sums the weights at that point (passing the summed value to the processor).

bryates:
> Thanks for this PR @bryates. I agree it'd be one way to fix the issue with the sum of the weights. However, I'm not sure if we want to do it in this way. Sometime is it useful to be able to easily process a smaller set of events. I think we should consider potentially adding something to the run script that checks if multiple JSONs correspond to the same sample, and just sums the weights at that point (passing the summed value to the processor).

That's an interesting point, but it might be easier to just remove a chuck of the root files when testing (or make some dedicated test json files), since the sum of weights shouldn't change (unless for some reason we wanted a different normalization for testing).

topcoffea/modules/combine_json_batch.py
```diff
@@ -0,0 +1,36 @@
...
+        return
+    nom = json.load(jsonFile) # Load the JSON information for the nominal file
+
+    for l in batch['files']:
+        nom['files'].append(l)
+
+    ''' Load the event counts, sum of weights, etc. '''
+    nom['nEvents']         = nom['nEvents']       + batch['nEvents']
+    nom['nGenEvents']      = nom['nGenEvents']    + batch['nGenEvents']
+    nom['nSumOfWeights']   = nom['nSumOfWeights'] + batch['nSumOfWeights']
```
kmohrman, line 31: This looks right to me, but for some reason the JSONs modified by this PR seem to have the same number for `nSumOfWeights` and `nGenEvents` (and it seems to correspond to `nGenEvents`). Not sure how that's happening if they were generated with this code though.

bryates: That's a good catch. I'm not sure why either, but I'll look into it.

bryates: I was going through the blame in the history of these files, and it looks like the last update was when @kmohrman used `update_json_sow.py`. I don't think this works anymore, since the script seemed to use the `SumOfWeights` histogram we've removed from the processor.

kmohrman: Hmm, so I think the `update_json_sow.py` stuff was written by @Andrew42 specifically to calculate the sum of the weights and put them in the JSON files, which is actually what allowed us to be able to drop the `SumOfWeights` hist (and that entire method of normalization) from the main processor to begin with. So, I'm not sure if the fact that the last update to the JSONs being the update to the sum of the weights could explain this particular issue.

Also, I think the `nSumOfWeights` listed in the JSONs prior to the merge is actually correct. I think it seems to be when the JSONs were merged together that the sum of the `nSumOfWeights` was apparently replaced by the sum of the `nGenEvents`. However, I could certainly be missing something.

bryates: I learned how to run the `sow` scripts, and added them to the README. The sub m sum of weights should now be correct.

kmohrman:
It looks like the 4t JSONs have been removed for UL17 and UL18. If this was done accidentally and manually, I guess we could just add them back in. However, if the script for combining the JSONs did this automatically, I guess we'd need to take a look at the logic and update it (so that this does not happen if we have to run this again). One thing to note is that the 4t samples from UL18 and UL17 are part of batch 4 (and not batches 1-3) so maybe that is what ended up confusing the logic of the script.

bryates:
> It looks like the 4t JSONs have been removed for UL17 and UL18. If this was done accidentally and manually, I guess we could just add them back in. However, if the script for combining the JSONs did this automatically, I guess we'd need to take a look at the logic and update it (so that this does not happen if we have to run this again). One thing to note is that the 4t samples from UL18 and UL17 are part of batch 4 (and not batches 1-3) so maybe that is what ended up confusing the logic of the script.

It looks like that's the issue. I've changed them to all have `b1` and am reprocessing now.

kmohrman:
@bryates, I'm not sure if changing b4->b1 is what we want to do. The 4t process was not included in b1 of the MC generation at all, so it could make it confusing to track down which samples are which (if we have b4 samples named as b1). I'm wondering if the logic we want might be something more like:
```
if there is more than one batch of the same sample:
    combine into one json (named according to the lowest batch number)
else:
    don't do anything
```

bryates:
> @bryates, I'm not sure if changing b4->b1 is what we want to do. The 4t process was not included in b1 of the MC generation at all, so it could make it confusing to track down which samples are which (if we have b4 samples named as b1). I'm wondering if the logic we want might be something more like:
> 
> ```
> if there is more than one batch of the same sample:
>     combine into one json (named according to the lowest batch number)
> else:
>     don't do anything
> ```

Ok, I've modified it to handle the `b4` correctly.

kmohrman:
Ok, I think all of the necessary updates have been made. I am currently running a test job over our private samples from all 4 UL periods, just to double check that the yields look consistent. Once that test is complete, I think we'll be ready to merge.

bryates:
> @bryates, have you actually tried processing any of these combined JSONs? Maybe I'm missing something obvious, but I'm getting an error that seems to be caused by some syntax error:
> 
> ```
> task id 1237 item pre_1237 failed: INPUT_MISS                                                                                                               
>     ('UL17_ttHJet_b1', '../../topcoffea/json/signal_samples/private_UL/UL17_tllq_b3.json/store/user/kmohrman/FullProduction/FullR2/UL17/Round1/Batch2/naodOnly_step/v3/nAOD_step_ttHJet_all22WCsStartPtCheckdim6TopMay20GST_run0/NAOD-00000_25306.root', 'Events') pre_1237 without result.
> ```
> 
> It seems the path to the json and the path to the root file are being combined together for some reason. I will check the logic in the run script to see where it's going wrong, but I'm just wondering if this is something on my end or if you've also seen it.

I was just seeing the same error. I'm not sure why this PR would cause that, but I'll look into it.

kmohrman:
Ok, my job (for all full R2 private samples) has completed, and I think the yields look reasonable:
```
ttx_multileptons-2lss_4t_m_2b.txt:observation 37.837
ttx_multileptons-2lss_4t_p_2b.txt:observation 48.097
ttx_multileptons-2lss_m_2b.txt:observation 316.731
ttx_multileptons-2lss_p_2b.txt:observation 441.646
ttx_multileptons-3l1b_m.txt:observation 86.089
ttx_multileptons-3l1b_p.txt:observation 106.317
ttx_multileptons-3l2b_m.txt:observation 82.413
ttx_multileptons-3l2b_p.txt:observation 102.115
ttx_multileptons-3l_sfz_1b.txt:observation 408.617
ttx_multileptons-3l_sfz_2b.txt:observation 319.427
ttx_multileptons-4l_2b.txt:observation 33.001
```
The total is 1982.29, which is I think consistent with what we'd expect (the numbers we've had before are quite close to this, though I don't have an exact number to compare against, since I've not rerun over the full samples since the latest SF PR, but at the very least this certainly does not look to be off by a factor of 3, which is what I'd expect if we messed up something with the sum of weights JSONs). 

So from my end, I'd be happy for this to be merged.

kmohrman (approved):
Just to reiterate my comment above, I think it looks like this is ready to be merged.

Issue #163, opened 2022-02-01
Systematics with NaN


bryates:
It looks like the `PreFire` systematics contain some `NaN` values, causing issues with the datacard maker:

```python
('PreFiringUp',) [ 0.00000000e+00  0.00000000e+00  7.05479544e-03  5.56694072e-02
  1.66547486e+05  1.39621340e+34  1.45597271e+16             nan
             nan             nan             nan             nan
             nan             nan             nan             nan
             nan             nan             nan             nan
             nan             nan             nan             nan
             nan             nan             nan             nan
  7.17334946e+31             nan             nan             nan
             nan  2.38377475e+29             nan             nan
             nan             nan             nan             nan
             nan             nan             nan             nan
  4.08705212e+32             nan  4.11227915e+16  4.65035655e+30
             nan             nan             nan  4.00895028e+24
...
```
For now we can try `np.nan_to_num()`, but we should figure out why. @andrea21z could this be related to the `nanoAODv8` vs `v9` differences?

andrea21z:
This is very strange, I have run with several samples for 2016, 2017 and 2018 with nanoAODv9 printing `events.L1PreFiringWeight.Up` and I don't see NaNs... And if you run with nanoAODv8 you will get an error because PreFiring branch does not exist, so I am not sure what is happening...

With which sample/year is this happening to you?

bryates:
> This is very strange, I have run with several samples for 2016, 2017 and 2018 with nanoAODv9 printing `events.L1PreFiringWeight.Up` and I don't see NaNs... And if you run with nanoAODv8 you will get an error because PreFiring branch does not exist, so I am not sure what is happening...
> 
> With which sample/year is this happening to you?

@kmohrman can you list the samples you ran to create `jan28_UL18SigBkgNoZZ_njets-bl0pt_withSys.pkl.gz`?

kmohrman:
As mentioned in PR #179, it turns out this issue was caused by the `add()` method of [`coffea.analysis_tools.Weights`](https://coffeateam.github.io/coffea/api/coffea.analysis_tools.Weights.html) modifying the up/down weights we pass to it. Our solution was to pass copies of the variables to `add()` (rather than the variables themselves).

PR #179, opened 2022-02-16
Fix bug where add() modifies up/down weights

kmohrman:
This PR fixes an issue related to the `add()` method of `coffea.analysis_tools.Weights`.  As mentioned in the [documentation](https://coffeateam.github.io/coffea/api/coffea.analysis_tools.Weights.html) this method can actually modify the up/down values that are passed. To avoid this issue, we're now passing copies of the variables instead of the variables themselves. For us, this issue was apparently only affecting the `L1PreFiringWeight` systematic (for unknown reasons). 

Unrelatedly, this PR also attempts to restore 4-space indents throughout the processor (as it had become a mix of 2-space indents and 4-space indents).  However, it does make it difficult to see what was actually changed in this PR, so please refer to [this commit](https://github.com/TopEFT/topcoffea/commit/4051d80b57a2a3e03133ffcb161cf7757819f21d) for the relevant `add()` changes.


bryates (approved):
Thanks @kmohrman. And this should close #163.

PR #180, opened 2022-02-16
Bugfix in flip estimation

sscruz:
Bug fix in the flip rate estimation, currently the transfer factor is set to 1 in most of the regions


bryates (approved):
Thanks @sscruz, seems simple enough.

kmohrman:
Looks good to me. Should we go ahead and merge it now?

PR #181, opened 2022-02-16
Update cfgs and jsons for skims with new lepMVA 

Andrew42:
This PR updates and makes available skims for the Run 2 datasets using NAODv9 and includes the lepMVA score for the new training. The lepMVA score is applied using the [`LepMVAULFriend.py`](https://github.com/sscruz/cmgtools-lite/blob/104X_dev_nano_lepMVA/TTHAnalysis/python/tools/nanoAOD/LepMVAULFriend.py) module provided by Sergio's [`104X_dev_nano_lepMVA`](https://github.com/sscruz/cmgtools-lite/tree/104X_dev_nano_lepMVA) branch of the [`cmgtools-lite`](https://github.com/CERN-PH-CMG/cmgtools-lite) repository.

As before, the skims were produced using the [LobsterSkimming](https://github.com/TopEFT/LobsterSkimming) code, which has also been updated to reflect changes needed to produce the skims. In particular, the lobster config switched from using the normal `Dataset` class to the `cmssw.Dataset` class. This means that the files processed came from querying DAS for the corresponding dataset, instead of directly from the json files. Another important note is that these skims now correctly include the data from the UL2016 MuonEG dataset.

Once the skims were produced, jsons for these samples were made by running the following command from the `topcoffea/analysis/topEFT` directory:
```
python make_skim_jsons.py --json-dir ../../topcoffea/json/data_samples --file ND_data_skim_locations.txt
```
The updated `ND_data_skim_locations.txt` file is also included in this PR.


bryates (approved):
Thanks @Andrew42, this looks good to me.

bryates:
@kmohrman and I discussed the other day, these will probably change the working points. Are there recommended points, or ROC curves available we can determine points from? Maybe that's a question for @sscruz and/or @barbaralvarez?

sscruz:
> @kmohrman and I discussed the other day, these will probably change the working points. Are there recommended points, or ROC curves available we can determine points from? Maybe that's a question for @sscruz and/or @barbaralvarez?

hi! this PR is only changing the cfgs that are being used, not the MVA variable itself. So it should be transparent to all results.
I'm working on a branch (https://github.com/TopEFT/topcoffea/tree/update-leptonmva) that actually implement those changes

bryates:
> > @kmohrman and I discussed the other day, these will probably change the working points. Are there recommended points, or ROC curves available we can determine points from? Maybe that's a question for @sscruz and/or @barbaralvarez?
> 
> hi! this PR is only changing the cfgs that are being used, not the MVA variable itself. So it should be transparent to all results. I'm working on a branch (https://github.com/TopEFT/topcoffea/tree/update-leptonmva) that actually implement those changes

Ah thanks! I mixed the two up in my head. So the question will apply to your updated lepMVA, but I guess you'll work on the ROC curves as you finish it up.

kmohrman:
> hi all,
> just one remark: we also need skims (or trees without any skim, as you prefer) with the new leptonMVA calculated for Monte Carlo. I think this PR only includes data, right?

Ah, that's a good point. If @Andrew42 plans to produce skims for the MC (signal and background), then I suppose this would be taken care of during the production of the skims (since I think @Andrew42 already added this to his skimming code he used for the data skims). However, if we do not end up skimming the MC (and I'm not sure if we will or not, since it generally does not take very long to process it), I'm wondering what would be the easiest and most straightforward way to get the new lepMVA included in the MC?

sscruz:
> > hi all,
> > just one remark: we also need skims (or trees without any skim, as you prefer) with the new leptonMVA calculated for Monte Carlo. I think this PR only includes data, right?
> 
> Ah, that's a good point. If @Andrew42 plans to produce skims for the MC (signal and background), then I suppose this would be taken care of during the production of the skims (since I think @Andrew42 already added this to his skimming code he used for the data skims). However, if we do not end up skimming the MC (and I'm not sure if we will or not, since it generally does not take very long to process it), I'm wondering what would be the easiest and most straightforward way to get the new lepMVA included in the MC?

I'm certainly fine with both approaches, however I would prefer to do skims (btw, I've already done them for 2016nonAPV, 2017 and 2018) because two reasons: we wouldn't need to write new code to update the MVA at nanoAOD production (it shouldn't be very complicated but also not straightforward as we would need to add era modifiers for 2016APV) and then we would get for free the fact that we would have a speed up when running the processor because we would be running over a smaller number of events

kmohrman:
> > > hi all,
> > > just one remark: we also need skims (or trees without any skim, as you prefer) with the new leptonMVA calculated for Monte Carlo. I think this PR only includes data, right?
> > 
> > 
> > Ah, that's a good point. If @Andrew42 plans to produce skims for the MC (signal and background), then I suppose this would be taken care of during the production of the skims (since I think @Andrew42 already added this to his skimming code he used for the data skims). However, if we do not end up skimming the MC (and I'm not sure if we will or not, since it generally does not take very long to process it), I'm wondering what would be the easiest and most straightforward way to get the new lepMVA included in the MC?
> 
> I'm certainly fine with both approaches, however I would prefer to do skims (btw, I've already done them for 2016nonAPV, 2017 and 2018) because two reasons: we wouldn't need to write new code to update the MVA at nanoAOD production (it shouldn't be very complicated but also not straightforward as we would need to add era modifiers for 2016APV) and then we would get for free the fact that we would have a speed up when running the processor because we would be running over a smaller number of events

Okay, that sounds good to me, though I'd also be interested in hearing @Andrew42's thoughts on this as well.

Andrew42:
I'm fine with making the skims for the MC, but I think we need to be careful about storing the weights. For example, we would obviously need to store the sum of the nominal event weights (evaluated at the SM), but also the up/down fluctuations on the MC PDF and fact./renom. uncertainties. I suppose those quantities could just be extracted later from the full samples, but would it make more sense to try and include them already in the skimmed samples?

kmohrman:
> I'm fine with making the skims for the MC, but I think we need to be careful about storing the weights. For example, we would obviously need to store the sum of the nominal event weights (evaluated at the SM), but also the up/down fluctuations on the MC PDF and fact./renom. uncertainties. I suppose those quantities could just be extracted later from the full samples, but would it make more sense to try and include them already in the skimmed samples?

@Andrew42, we're already working on this issue for the factorization/renormalization and ISR/FSR up/down weights. The current plan is to handle it with a similar method to the `sow_processor`. We would process the full (unskimmed) samples to extract the up/down weights (in principle would only need to be done once), then just store those sum of weights values in the JSON. Then, when running the analysis processor (over skimmed or unskimmed samples), we can just grab the values from the JSON. So, I I think this issue can be handled completely independently of the skimming, right?

Andrew42:
@kmohrman that's a good point. Since we already have a mechanism in place (`sow_processor`), I think it makes the most sense to not worry about including the sow for the up/down fluctuations in the skims and include them by extending the `sow_processor` instead, which would eventually result in them showing up in the json files instead. This way the normalizations would all be handled in a consistent way. As far as actually generating the MC skims, I'll get started on the signal samples, but will include them in a different PR.

Andrew42:
The most recent commit updates the JSON files to point to the newest set of data skims. These new skims were produced by running over the data copied to the ND_T3 and should include Sergio's fix to the lepMVA module that caused rare crashes. As the changes are only applied to the JSON files, no changes should be required if you're using the `data_samples_NDSkim.cfg` config file

analysis/topEFT/work_queue_run.py
```diff
@@ -182,8 +182,8 @@ def LoadJsonToSampleName(jsonFile, prefix):
 
 executor_args = {
     'master_name': '{}-workqueue-coffea'.format(os.environ['USER']),
-
-    'xrootdtimeout': 180,
+    'filepath':'/tmp',
+    'xrootdtimeout': 300,
```
kmohrman, line 186: @Andrew42, I think I'd prefer to keep the same default settings for the timeout and filepath committed for now. Is there a reason why you wanted to change these? Or were they maybe just accidentally committed with other updates?

analysis/topEFT/work_queue_run.py
```diff
@@ -198,7 +198,7 @@ def LoadJsonToSampleName(jsonFile, prefix):
     'schema': NanoAODSchema,
     'skipbadfiles': False,
 
-    'retries': 5,
+    'retries': 10,
```
kmohrman, line 201: Same comment as above.

topcoffea/cfg/data_samples.cfg
```diff
@@ -88,23 +88,23 @@ root://ndcms.crc.nd.edu/
...
+#
+#../../topcoffea/json/data_samples/2018/MuonEG_A-UL2018.json
+#../../topcoffea/json/data_samples/2018/MuonEG_B-UL2018.json
+#../../topcoffea/json/data_samples/2018/MuonEG_C-UL2018.json
+#../../topcoffea/json/data_samples/2018/MuonEG_D-UL2018.json
+#
+#../../topcoffea/json/data_samples/2018/SingleMuon_A-UL2018.json
+#../../topcoffea/json/data_samples/2018/SingleMuon_B-UL2018.json
+#../../topcoffea/json/data_samples/2018/SingleMuon_C-UL2018.json
+#../../topcoffea/json/data_samples/2018/SingleMuon_D-UL2018.json
```
kmohrman, line 110: Just wondering why these JSONs are now commented by default? Was that intentional?

topcoffea/cfg/data_samples_NDSkim.cfg
```diff
@@ -41,39 +41,48 @@ root://deepthought.crc.nd.edu/
...
+#../../topcoffea/json/data_samples/2017/SingleElectron_C-UL2017_NDSkim.json
+#../../topcoffea/json/data_samples/2017/SingleElectron_D-UL2017_NDSkim.json
+#../../topcoffea/json/data_samples/2017/SingleElectron_E-UL2017_NDSkim.json
+#../../topcoffea/json/data_samples/2017/SingleElectron_F-UL2017_NDSkim.json
+#
+#../../topcoffea/json/data_samples/2017/SingleMuon_B-UL2017_NDSkim.json
+#../../topcoffea/json/data_samples/2017/SingleMuon_C-UL2017_NDSkim.json
+#../../topcoffea/json/data_samples/2017/SingleMuon_D-UL2017_NDSkim.json
+#../../topcoffea/json/data_samples/2017/SingleMuon_E-UL2017_NDSkim.json
+#../../topcoffea/json/data_samples/2017/SingleMuon_F-UL2017_NDSkim.json
```
kmohrman, line 85: Same comment as above: why are we commenting these JSONs? I think I'd prefer to leave all of the Run 2 JSONs uncommented, since we'll most often be running over the full Run 2 together.

analysis/topEFT/work_queue_run.py
Andrew42: fixed with 80bbed74e99fe51bc931a15c9f7ddaec83a66c67

analysis/topEFT/work_queue_run.py
Andrew42: fixed with 80bbed74e99fe51bc931a15c9f7ddaec83a66c67

topcoffea/cfg/data_samples.cfg
Andrew42: fixed with 80bbed74e99fe51bc931a15c9f7ddaec83a66c67

topcoffea/cfg/data_samples_NDSkim.cfg
Andrew42: fixed with 80bbed74e99fe51bc931a15c9f7ddaec83a66c67

kmohrman:
Thanks for reverting those changes @Andrew42. It looks good to me, I'd be happy for it to be merged at this point.

PR #186, opened 2022-02-18
Trigger sf

andrea21z:
-> Updating trigger SFs including new lep triggers.
-> Fixing the way to calculate trigger SFs unc.
-> Fixing muon reco SFs (Only available at low pT).


topcoffea/modules/corrections.py
```diff
@@ -454,14 +455,14 @@ def GetTriggerSF(year, events, lep0, lep1):
   ls=[]
   for i in [0,1,2]:
```
bryates, line 456: What is this `i` doing exactly? Can you add a comment, or a descriptive variable name?

bryates (commented):
Thanks @andrea21z, I just had the one question.

topcoffea/modules/corrections.py
andrea21z: `i` is just an index: i=0 for nominal SF, i=1 for SFdown and i=2 for SFup

bryates: Could you change `i` to something like `syst` or `shift` and then put a comment for the nominal/up/down numbers?

andrea21z: sure!

bryates: Thanks!

analysis/topEFT/topeft.py
```diff
@@ -332,7 +332,7 @@ def process(self, events):
           
             # Trigger SF
             GetTriggerSF(year,events,l0,l1)
-
+
```
kmohrman, line 335: Also, this is not very important, but could you remove this added whitespace? Other than just being unnecessary, it might also make the PR slightly less clear if we look back at this in the future (since I think this PR does not need to change the processor at all, but at first glance we might think that it does until we look more closely and see that it was just whitespace that was added).

kmohrman:
Thanks for all of the updates @andrea21z! This PR now looks good to me, so if @bryates agrees, I'd be happy for it to be merged.

PR #190, opened 2022-02-21
Pin python to 3.9

bryates:
Pin Python to 3.9 to fix the work queue CI.


bryates:
Thanks for your changes @kmohrman, looks like everything is passing now.

kmohrman:
Actually I still think something might be wrong...

tests/test_futures.py
```diff
@@ -29,34 +29,6 @@ def test_nonprompt():
...
-        "analysis/topEFT/test/UL17_private_ttH_for_CI_yields.json",
-        "-t1",
-        "New yields",
-        "-t2",
-        "Ref yields"
-    ]
-
-    # Run comparison
-    subprocess.run(args)
-
```
bryates, line 59: Maybe we should keep these after all, since the code coverage goes down a bit without them. I agree it's a bit redundant, but the prior test should halt the CI if things fail.

kmohrman:
> Actually I still think something might be wrong...

I think that the differences in yields between CI and ref (~1%) are not expected. I'm wondering if maybe there was something uncommitted in your env when you updated the ref? When I remade the ref in my env it seems to agree with what the CI is giving. I will try pushing that and see if the comparison agrees more closely

tests/test_futures.py
klannon: Is it just that the code written to compare the yields (in other words the code that used only in this test and nowhere else) isn't covered?  That's not a good reason to keep it in the tests.  Instead you should remove the unused code from the repo.

bryates: That's a fair point. I think some of it might be used elsewhere, but mostly for extracting yields.

kmohrman: @klannon and @bryates, sorry if I'm just misunderstanding your points, but I think this code is definitely used, and definitely not something we want to get rid of. The code in question consists of the two scripts that are responsible for taking a pkl and getting the yields in JSON format, and for comparing the yields in two JSONs. These scripts are explicitly used in a later step of the CI to test check the output pkl file produced by the `test_futures` job. 

Previously, these scripts had been getting run in the `test_futures` as well. However, I'm not sure this was very useful, as the job was not even set up to fail if the yields disagreed, plus we have a separated, dedicated job for creating and comparing the yields anyway, so it seemed redundant (and potentially confusion-prone) to have it in the futures test as well. For these reasons, I just removed the yields checks from test_futures. 

However, I am not sure why that made the coverage go down, except for the fact that the dedicated jobs that create and compare the yields are just run directly in as jobs in the yaml file ([here](https://github.com/TopEFT/topcoffea/blob/python39/.github/workflows/main.yml#L66-L74)) i.e. not through `pytest`, so maybe they're not being tracked by codecov.

klannon: > except for the fact that the dedicated jobs that create and compare the yields are just run directly in as jobs in the yaml file ([here](https://github.com/TopEFT/topcoffea/blob/python39/.github/workflows/main.yml#L66-L74)) i.e. not through `pytest`

That would explain it.

kmohrman: Ok, so now that we understand why the codecov changed, @bryates I'm wondering if you still object to the removal of the yields test from the `test_futures` job?

kmohrman:
Ok @bryates I think we've made all of the relevant updates for this branch (including updating the ref yields for the `triggerSF` updates, which we should have done on PR #186, but we missed because of our misleading CI issue discussed [here](https://mattermost.web.cern.ch/topeft-runii/pl/tusbyug6k3nr9qwqzfan9srg5w), and on this branch we've also updated the CI so that this issue should no longer happen). 

Were there any other changes or updates you wanted to make before we merge?

bryates:
> Ok @bryates I think we've made all of the relevant updates for this branch (including updating the ref yields for the `triggerSF` updates, which we should have done on PR #186, but we missed because of our misleading CI issue discussed [here](https://mattermost.web.cern.ch/topeft-runii/pl/tusbyug6k3nr9qwqzfan9srg5w), but on this branch we've also updated the CI so that this issue should no longer happen).
> 
> Were there any other changes or updates you wanted to make before we merge?

I don't have anything else. Once the CI is done, I'll merge it.

bryates:
> > Ok @bryates I think we've made all of the relevant updates for this branch (including updating the ref yields for the `triggerSF` updates, which we should have done on PR #186, but we missed because of our misleading CI issue discussed [here](https://mattermost.web.cern.ch/topeft-runii/pl/tusbyug6k3nr9qwqzfan9srg5w), but on this branch we've also updated the CI so that this issue should no longer happen).
> > Were there any other changes or updates you wanted to make before we merge?
> 
> I don't have anything else. Once the CI is done, I'll merge it.

Looks like the work queue failed at the datacard maker stage. I'll look into it tomorrow.

kmohrman:
> > > Ok @bryates I think we've made all of the relevant updates for this branch (including updating the ref yields for the `triggerSF` updates, which we should have done on PR #186, but we missed because of our misleading CI issue discussed [here](https://mattermost.web.cern.ch/topeft-runii/pl/tusbyug6k3nr9qwqzfan9srg5w), but on this branch we've also updated the CI so that this issue should no longer happen).
> > > Were there any other changes or updates you wanted to make before we merge?
> > 
> > 
> > I don't have anything else. Once the CI is done, I'll merge it.
> 
> Looks like the work queue failed at the datacard maker stage. I'll look into it tomorrow.

Sorry, I guess I must be missing something. Where do you see the failure? To me it seems to look like the whole CI is passing.

bryates:
> > > > Ok @bryates I think we've made all of the relevant updates for this branch (including updating the ref yields for the `triggerSF` updates, which we should have done on PR #186, but we missed because of our misleading CI issue discussed [here](https://mattermost.web.cern.ch/topeft-runii/pl/tusbyug6k3nr9qwqzfan9srg5w), but on this branch we've also updated the CI so that this issue should no longer happen).
> > > > Were there any other changes or updates you wanted to make before we merge?
> > > 
> > > 
> > > I don't have anything else. Once the CI is done, I'll merge it.
> > 
> > 
> > Looks like the work queue failed at the datacard maker stage. I'll look into it tomorrow.
> 
> Sorry, I guess I must be missing something. Where do you see the failure? To me it seems to look like the whole CI is passing.

Ok, good. I got an email yesterday, and the timing was so close I had just assumed it was for this PR, but it was another branch failling.

tests/test_futures.py
kmohrman: Ok, but I guess I'm still not completely sure hat I understand why we need to check the yields here. We have dedicated jobs for that later in the CI. To me, having the "run the processor" check separate from the "check the yields" check makes more sense since that way it's easier to see what's wrong if something fails. The way it's set up now (after you added the yields checks back into the futures job), if anything is wrong with the processor or if something unexpected changed that affects the yields, this job will fail (and that will make it harder to figure out what went wrong, and we'll have to spend more time digging through the CI output trying to find where the actual error came from).

bryates: Let me think a bit about how we can keep the code coverage for these files, while avoiding what you described.

bryates: @kmohrman would you be opposed to replacing these lines
https://github.com/TopEFT/topcoffea/blob/0e5642fe781386acf632542bf81aa6086f5cf4d4/.github/workflows/main.yml#L66-L74
with the `pytest` versions? Of course making sure to point to the `json` files from the "run the processor" section.

kmohrman: Sure, that sounds reasonable to me. Though, just to make sure I'm following, when you say "point to the json files from the "run the processor" section", do you mean pkl files? Or am I missing something?

bryates: Yes, sorry, point it to the `pkl` file.

kmohrman: Ok, sounds good to me. Though, just to check, did you actually push the new purest files for the test_compare_yields_after_processor and test_make_yields_after_processor checks? I might be missing them, but don't see them in the list of changed files. Also, in the yaml file, should it be something like `tests/test_make_yields_after_processor.py` instead of just `test_make_yields_after_processor`?

bryates: Thanks for catching that! I just added the new file.

bryates: For the names, `pytest` looks for everything in the `tests` folder, and any python file _or_ any function name defined in any of them is able to be called.

bryates: I should clarify, the `-k` flag allows this, which I had forgotten to add.

PR #191, opened 2022-02-22
Move calls to Weights add() outside of loops

kmohrman:
This PR moves some of the calls to the `add()` function (of the `analysis_tools.Weights()` object) out of some of the processor's for loops (over systematics [here](https://github.com/TopEFT/topcoffea/blob/1e71a2a3034f18ca7d2b4df89f187ba4b24d6398/analysis/topEFT/topeft.py#L211), and over lepton categories [here](https://github.com/TopEFT/topcoffea/blob/1e71a2a3034f18ca7d2b4df89f187ba4b24d6398/analysis/topEFT/topeft.py#L365)). For the calls that can be moved out of all loops, I've removed the `copy.deepcopy()`, as it should no longer matter if we're modifying the up/down weights (since they'll never be used again). However, some of the calls to `add()` must remain inside of these loops, since the systematics' weights depends on the loop variable; in these cases, we need to leave the `copy.deepcopy()` to ensure that the up/down variables are not modified.

There is of course a concern that this rearrangement could have introduced bugs. The updates pass the CI, meaning that the nominal values remain consistent. To check all of the up/down variations of all of the systematics, I've run the processor (with systematics turned on) over a set of example samples (ttH samples from all 4 UL periods) with and without these updates to the processor. Comparing the yields of the nominal and up/down hists, the largest absolute difference (after normalizing) was about 0.003, and the largest percent difference was about 0.2%. I think these differences could be explained by rounding errors (potentially exacerbated by the randomness introduced by the Rochester SFs), but if anyone has any objections or would suggest further tests and checks, just let me know.

Also, as an additional verification, I'm wondering if @andrea21z you could confirm that [these systematics](https://github.com/TopEFT/topcoffea/blob/1e71a2a3034f18ca7d2b4df89f187ba4b24d6398/analysis/topEFT/topeft.py#L196-L204) that I've moved out of all loops in fact do not depend on any of the loop variables? Thanks!


bryates (approved):
Thanks @kmohrman, I'm happy with your comment about the yields not changing (modulo rounding and shifts from the random numbers).

kmohrman:
Thanks @bryates and @andrea21z. I just realized something about these changes, I think there might be an issue in the way I updated the `isData` if statements in the inner for loop. Originally, the `lepSF` and `FF` were applied to both MC and data. Just to confirm, is that correct @andrea21z? With the updates in this PR, it looks like I've accidentally only applied these to MC (which was not caught in my tests as I was only looking at MC, not data).

andrea21z:
> Thanks @bryates and @andrea21z. I just realized something about these changes, I think there might be an issue in the way I updated the `isData` if statements in the inner for loop. Originally, the `lepSF` and `FF` were applied to both MC and data. Just to confirm, is that correct @andrea21z? With the updates in this PR, it looks like I've accidentally only applied these to MC (which was not caught in my tests as I was only looking at MC, not data).

This is a good catch. Lepton SFs have to be applied only in MC! (I guess the indentation was changed by mistake at some point) And I think the fakefactor has to be applied only in data... but maybe @sscruz can comment here.

sscruz:
hi indeed `SF` have to be applied only to data, however that's not the case for the fake-factor. The `FF` should be applied also to MC because the estimation is `FF*(data_application_region - MC_application_region)`. The `FF` is 1 by construction for events in the signal region so this should not affect the MC predictions in the signal region, only the fakes estimate.

sscruz:
> Ok, thanks @sscruz. So I think your explanation makes sense for the `FF`. I will update this branch so that it's applied to both data and MC (as it is on the master). However, I just want to double check what you said about the `lepSF`:
> 
> > indeed SF have to be applied only to data
> 
> Just to confirm you meant to write "data" here (or in case it was a typo and meant to be "MC")? Just want to double check since @andrea21z mentioned that she thought it should be for MC, and I would have guessed that as well (since in general SFs are used to make the MC better match the data, right?)

sorry, I wrote too quickly :) I meant SF should be applied only to MC

kmohrman:
> > Ok, thanks @sscruz. So I think your explanation makes sense for the `FF`. I will update this branch so that it's applied to both data and MC (as it is on the master). However, I just want to double check what you said about the `lepSF`:
> > > indeed SF have to be applied only to data
> > 
> > 
> > Just to confirm you meant to write "data" here (or in case it was a typo and meant to be "MC")? Just want to double check since @andrea21z mentioned that she thought it should be for MC, and I would have guessed that as well (since in general SFs are used to make the MC better match the data, right?)
> 
> sorry, I wrote too quickly :) I meant SF should be applied only to MC

Okay, thank you! I'll make that update here as well.

kmohrman:
Ok, just finished double checking the yields after the commit above, and the the MC still looks consistent with the master branch. I also tried running over data, and that looks consistent with the master branch as well. This was surprising to me at first, since on the master branch, the `lepSF` also looks like it is being applied to the data (which, as we discussed above, would be incorrect). However, I think we might actually protect against this [here](https://github.com/TopEFT/topcoffea/blob/06141f2c01ca4135f7f2f0f43fc2b06ccfc32c69/analysis/topEFT/topeft.py#L698), where we take only the FF and flips weights in the case of data. Nevertheless, I don't see any reason why we would want to add these weights to the `Weights` object for data when we're never going to use it, so I think the updates we made on this branch (that now only add the FF and fake weights for data) make sense.

I'm wondering if anyone has any other comments or if we should proceed with merging this PR?

PR #193, opened 2022-02-26
Add Private MC Skims

Andrew42:
This PR adds the cfg and json files that point to skims for the Full Run2 private MC signal samples. As in PR #181, these skims should include the lepMVA weights in the skimmed NAOD.

As usual, the command used to generate the json files was:
```
python make_skim_jsons.py --json-dir ../../topcoffea/json/signal_samples/private_UL --file ND_private_mc_skim_locations.txt
```
The new `ND_private_mc_skim_locations.txt` and a corresponding cfg file, `mc_signal_samples_NDSkim.cfg`, are also included in this PR.


bryates (commented):
Thanks @Andrew42. Will @abasnet97 be repeating the exercise to validate these?

I'll hold off on approving this until that's done.

abasnet97:
> Thanks @Andrew42. Will @abasnet97 be repeating the exercise to validate these?
> 
> I'll hold off on approving this until that's done.

Hi @bryates, yes. I will start working on validating these this afternoon.

Andrew42:
I've updated the ND private MC jsons to point to the new skims that fix the crash related to evaluating the lepMVA variables dxy and dz. @abasnet97 for what concerns your validation checks, you can continue using the same `cfg` file and you should simply have to do a `git pull` to automatically start using the new skims.

bryates:
> As discussed at the meeting last Friday as well as on mattermost, it might be beneficial to merge this branch, even though the skims have not been fully validated yet (since while we definitely want to track down the strange percent level discrepancies, we can still use the skims to move forward in other areas of the analysis while the differences are being tracked down).
> 
> For those reasons, I think we could merge this branch now, if @bryates agrees.

I agree, let's merge this PR.

PR #194, opened 2022-03-01
Sum of weights for PS and scale systs

kmohrman:
This PR adds functionality for handling the normalization of the ISR/FSR and renormalization/factorization systematics.

For these systematics, the denominator in the normalization for the up (or down) histograms should be the sum of the weights for all events with the up (or down) fluctuation (_not_ the sum of the nominal weights).  In order to make this information accessible to the processor, we have decided to store this information in the JSON files for the samples. Summary of changes:
* Modified the `sow_processor.py` and the `update_json_sow.py` script to also calculate the sum of the PS and scale up/down weights (and be able to update the JSONs accordingly)
* Ran these scripts over our private samples to calculate the weights and update the JSONs
* Modified the run scripts (for both futures and WQ) to grab the info from the JSON and pass it to the processor
* Modified the processor to scale the PS and scale up/down hists with the appropriate sum of weights

In order to check that these changes were working as expected, I ran a test run with and without the changes to the processor (with the systematics turned on), and compared the resulting histograms. As expected, the nominal and all of the up/down hists (besides the PS and scale hists) were consistent before and after the updates; only the PS and scale hists showed significant changes. Plotting at the histograms, the up/down variations for the PS and scale hists were significantly smaller after the updates, as expected. Just to double check that these changes did not introduce any bugs into other parts of the code, I also ran a quick check on a few data samples (to make sure the yields in the CRs exactly agreed before and after the changes, and they did).

Some open items for discussion:
* The changes implemented so far are only for our private samples, I have not processed our central background samples with these updates (mainly because I ran into some ndcms issues when I attempted to run). So, if we were to merge this PR in its current state, it would actually break our ability to process our background samples (since the code is expecting the sum of the up/down weights to be stored in the JSON for all MC samples, and they are not for the central samples). However, while the review of this PR is taking place, I can attempt to process the central samples again, and will update the PR if I am successful; if I am unsuccessful, I can update the code to handle the crash in a more informative way (e.g. add some exception about missing weights).
* There is also an additional consideration brought up by @rgoldouz; it seems that for some central samples, the length of the vector that stores the scale weights is actually of length 8 rather than 9 (as the nominal weights are not included). If we can manage to process some central samples, I can add a check to `corrections.py` to handle this. I should emphasize, however, that I do not think this issue is affecting any of the sum of weights that have been calculated so far for the private samples (since we explicitly ask for the 8th index, I think we would have expected an out of range error had this issue been affecting our samples).


analysis/topEFT/update_json_sow.py
```diff
@@ -28,10 +28,50 @@
...
+    'renormfactDown',
+]
+
+# Construct a dict to hold the hist name and json name, format:
+# d = {
+#   'varUp': {
+#       'hist_name': 'SumOfWeights_varUp',
+#       'jsn_key_name': 'nSumOfWeights_varUp',
+#   }
+# }
```
bryates, line 53: Can we remove this commented out code?

bryates (approved):
Thanks @kmohrman, I think overall this PR looks good. Can you post (or link to) the plots you described showing how things like ISR/FSR change?

analysis/topEFT/update_json_sow.py
kmohrman: This comment is just supposed to explain the naming convention being used in the dictionary that it is constructing (since it might not be obvious otherwise). However, if you'd have a suggestion about how to express this in a more useful way I'd of course be happy to modify the comment.

bryates: Oh sorry, I guess I scrolled to fast and thought it was actual code. This is actually a good thing to keep in!

kmohrman: Also, just to try to clarify, this is not actually commented out code, it's really just pseudo code that's supposed to explain for the structure of the dictionary that the function creates.

kmohrman:
> Thanks @kmohrman, I think overall this PR looks good. Can you post (or link to) the plots you described showing how things like ISR/FSR change?

Sure, I've attached a before and after plot (with just PS and scale systematics included). The first plot is before the updates, the second plot is after. This example is for just ttH UL18 in the 3l category. The distribution is `lj0pt`. 
image
image

bryates:
> > Thanks @kmohrman, I think overall this PR looks good. Can you post (or link to) the plots you described showing how things like ISR/FSR change?
> 
> Sure, I've attached a before and after plot (with just PS and scale systematics included). The first plot is before the updates, the second plot is after. This example is for just ttH UL18 in the 3l category. The distribution is `lj0pt`. image image

Thanks! That's really nice to see how little we are affected by the shape uncertainties.
Comparing to a fractional change plot from the TOP-19-001 AN
image
most of the systematics were `O(<1%)`, so it looks like this agrees. Maybe we can eventually make similar plots, but it doesn't need to hold up this PR.

kmohrman:
Thanks @bryates. Regarding the other open items, I've implemented a check ([this commit](https://github.com/TopEFT/topcoffea/pull/194/commits/c9691d68dcbb2fe0b6ff66e47f942c347b0335bb)) for the 8 vs 9 LHE scale weights issue @rgoldouz mentioned. I'm now attempting to process the backgrounds again (`ndcms` seems to be cooperating so far) so I'll plan to update the background JSONs with the proper sum of weights if the job manages to finish. If not, I guess we could consider proceeding without it till we're actually able to process the backgrounds.

Also, I'm wondering if @Andrew42 would have any comments on this PR, since he was the one who handled the sum of weights issue for these systematics for TOP-19-001?

kmohrman:
Ok, I think we might finally be ready to merge this PR. Just to summarize the latest issue for future reference:
* We noticed that some of the scale weights were 0 for one the the central samples, and added a check to convert the weights to 1 in the case of 0s
* However, as explained by @sscruz the weights of 0 are probably not unexpected
* We thus removed the check for 0 weights (so now we just leave them alone and do not attempt to set them to 1).

One important thing to note about merging this PR is that it will break our ability to process background samples with the systematics option turned on. This is because when running with systematics, the code is expecting to find the sum of weights for the up/down variations for the scale and PS systematics, but we cannot currently calculate these for the background samples (because of `ndcms` issues). However, since we can't currently process the backgrounds (with or without systematics) I think the fact that we in principle cannot run with systematics when processing background samples will not matter at this point. For this reason, I think we should proceed to merge this branch and we should update the background sample JSONs later when we are someday (hopefully) able to process backgrounds again. I've created an Issue #196  so that we remember to do this.

@bryates (or anyone else), any final comments, or should we proceed with the merge?

bryates:
> Ok, I think we might finally be ready to merge this PR. Just to summarize the latest issue for future reference:
> 
> * We noticed that some of the scale weights were 0 for one the the central samples, and added a check to convert the weights to 1 in the case of 0s
> * However, as explained by @sscruz the weights of 0 are probably not unexpected
> * We thus removed the check for 0 weights (so now we just leave them alone and do not attempt to set them to 1).
> 
> One important thing to note about merging this PR is that it will break our ability to process background samples with the systematics option turned on. This is because when running with systematics, the code is expecting to find the sum of weights for the up/down variations for the scale and PS systematics, but we cannot currently calculate these for the background samples (because of `ndcms` issues). However, since we can't currently process the backgrounds (with or without systematics) I think the fact that we in principle cannot run with systematics when processing background samples will not matter at this point. For this reason, I think we should proceed to merge this branch and we should update the background sample JSONs later when we are someday (hopefully) able to process backgrounds again. I've created an Issue #196 so that we remember to do this.
> 
> @bryates (or anyone else), any final comments, or should we proceed with the merge (once the CI finishes)?

I have no further comments. Feel free to merge.

PR #198, opened 2022-03-08
Fix pileup reweighting

sscruz:
Data and MC distributions in the pile-up reweighting are not properly normalized in 2016 due to the fact that we separate APV and non-APV. The fix changes the pile-up distribution for data, so the proper pile-up profile of each era is used, but also we are forcing the normalization of the samples to be 1.


kmohrman:
Thanks very much for this fix @sscruz. Looks like the CI is failing, and it looks like it might be the comparison of the datacards. However, I'm not sure that this would be expected, since I think the CI only uses a 2017 file for all of the tests and comparisons.

sscruz:
hi, the change should have a minor impact in 2017 and 2018 as well. In this part of the code [1], I'm forcing data and MC pile-up distributions to be normalized. The normalization was close to be the same for 2017 and 2018, but not exactly, so one can expect minor differences (I read 0.18% differences in some of the log, I think this is reasonable) 

[1] https://github.com/TopEFT/topcoffea/pull/198/files#diff-64dfefa7d075e64c7534364b92b5c722612d80f0788662bae96706794369ddcdL246-L255

kmohrman:
Ah, okay. Thanks for explaining that. Just one quick followup, sorry if this is obvious, but I'm wondering what it means for the MC pile-up distribution to be normalized (normalized to what)?

Alos, since these changes are expected, I'm wondering if you could run the `remake_ci_ref_datacard.py` and `remake_ci_ref_datacard.sh` scripts so that the CI will pass before we merge?

bryates (commented):
Thanks for fixing this @sscruz! Can you update the CI references? Then we can get merge this ASAP.

bryates (approved):
Thanks for updating the CI @sscruz. I've approved this PR, but @kmohrman is updating the other yields not used in the CI, so I'll hold off on merging for now.

kmohrman:
Ok, just updated the ref yields we use for the larger-scale yields tests. This test is also just with 2017 samples, and consistent we already saw from the CI ref yields that @sscruz updated, the percent differences are fairly small (around O(0.1%) or so). I'd be happy for the PR to be merged at this point.

Issue #79, opened 2021-08-18
Possible issue for 2016APV with filters in event selection?


kmohrman:
In the `events.Flag` filters in the event selection (in `topcoffea/modules/selection.py`), there are several if statements that evaluate whether the year is 2016. What should happen if the year is 2016APV? Right now, I think the statement would evaluate the same way for 2016APV as it does for 2017 and 2018 (since `"2016APV" != "2016"`). Is this behavior correct?

kmohrman:
Closed with PR #202.

PR #202, opened 2022-03-22
Update central bkg samples

kmohrman:
In this PR:
* Updates to central background JSONs for new v9 smples
* Updates to the `make_json` script for the new v9 background samples
* Bug fix for `sow_processor` for up/down weights for central samples (the up/down weights should be scaled by the nominal gen weight before summing, not just summed directly)
* Update of central background JSONs to include the up/down weights from the `sow_processor`
* Unrelated bug fix to address Issue #79 (since @sscruz confirmed on [mattermost](https://mattermost.web.cern.ch/topeft-runii/pl/k4hdkrd8qbycmc78ot8gjyq7uw) that this should apply to both 16 and 16APV).


bryates (approved):
Thanks @kmohrman, this looks good!

PR #205, opened 2022-03-23
Switch private MC skims produced with correct lepMVA module

Andrew42:
This PR is to rectify a mistake made in PR #193, which added skims of the private MC samples. I realized that my logic for selecting the correct lepMVA module was not working properly when running over the private samples. This bug did not affect the data skims, but which still need to be remade anyways due to the dxy/dz error.

`ND_private_mc_skim_locations.txt` has been updated to point to the location of the remade skims and the JSON files were all updated using:
```
python make_skim_jsons.py --json-dir ../../topcoffea/json/signal_samples/private_UL --file ND_private_mc_skim_locations.txt
```
No changes are needed for any of the existing `cfg` files as this doesn't add or remove any JSONs from the repo, only modifies already existing ones.


kmohrman:
Thanks @Andrew42, this looks good to me. Unless anyone else objects, I'd be happy for it to be merged.

bryates (approved):
Looks good to me as well.

PR #208, opened 2022-03-24
Update data jsons with most up-to-date samples

Andrew42:
The primary purpose of this PR is to update the [data JSON](https://github.com/TopEFT/topcoffea/tree/update-data-jsons/topcoffea/json/data_samples) files after it was noticed that the currently used versions have been invalidated. As done previously, the datasets used are taken from the PdmV [twiki page](https://twiki.cern.ch/twiki/bin/viewauth/CMS/PdmVRun2LegacyAnalysis) for UL Run2 analyses.

The datasets which were updated almost entirely consisted of UL2016 "pre-VFP" samples, except for [DoubleMuon_F-UL2017.json](https://github.com/TopEFT/topcoffea/commit/83e46968ea677cd971d6882cff0cac230bb5abd2#diff-fdb93a58f3c4e6f5c207c4f6c3aa54f8da035d814b14632d8f009a33350b23b9), where apparently the old JSON was missing a single root file. This is slightly concerning as I'm not entirely sure why that root file wasn't included originally, but seeing as how all the other re-created 2017 and 2018 JSON files matched perfectly, I think it's probably not something on our end.

This PR also introduces minor modification to [createJSON.py](https://github.com/TopEFT/topcoffea/blob/update-data-jsons/topcoffea/modules/createJSON.py#L147-L150) that skips trying to open the first root file from a dataset in order to determine the list of WCs if that dataset is being queried from DAS. The reason being, none of the EFT samples are available through DAS and so it doesn't make sense to open an xrootd connection to this remote file when we already know it won't have a list of WCs. Aside from a slight performance speed-up, this change circumvents transient file read errors related to accessing the file via xrootd.


bryates (approved):
Thanks @Andrew42, looks good.

kmohrman:
Looks good to me as well. Thanks @Andrew42.

PR #207, opened 2022-03-24
Fix ext samples

bryates:
This PR updates some samples (`topcoffea/json/background_samples/central_UL/UL16APV_WZZ.json`, `topcoffea/json/background_samples/central_UL/UL16APV_ZZZ.json`, `topcoffea/json/background_samples/central_UL/UL16_WZZ.json`, and `topcoffea/json/background_samples/central_UL/UL16_ZZZ.json`) to fold the extensions into the nominal json files.

Issue #99, opened 2021-09-16
Update background samples once UL become available


kmohrman:
When we created the JSONs for the background samples, not all processes had a UL sample available. These processes are marked with a `NOTE: PLACEHOLDER` comment in the `make_json.py` file. Once the UL samples become available, we should replace these placeholders with the appropriate samples.

kmohrman:
Only remaining non-v9 background samples:
* UL18_TTJets
* UL16APV_ZZTo4L

Currently using these dasgo queries to look for the missing samples:
```
dasgoclient -query "dataset=/TTJets_TuneCP5_13TeV-amcatnloFXFX-pythia8/*v9*UL*/NANOAODSIM"
```
```
dasgoclient -query "dataset=/ZZTo4L_TuneCP5_13TeV_powheg_pythia8/*v9**UL*/NANOAODSIM"
```

kmohrman:
Closed by PR #217.

PR #211, opened 2022-03-28
Add mc background skim samples

Andrew42:
This PR adds a cfg and the associated json files for the skimmed MC background samples. The samples were produced using the usual LobsterSkimming workflow and should reflect the changes introduced by PR #207, which merged the files from a number of `*_ext` jsons into their non-ext counterparts.

The command to produce the json files is given below:
```
python make_skim_jsons.py --json-dir ../../topcoffea/json/background_samples/central_UL --file ND_bkgd_mc_skim_locations.txt
```
An important thing to note is that the background samples `UL16APV_ZZTo4L` and `UL18_TTJets` were not included in this set of skims as they still do not have NAODv9 versions available. I will make a git issue detailing this to help make sure we don't forget.


kmohrman:
Thanks @Andrew42! Regarding the issue, you might want to just post on [this one](https://github.com/TopEFT/topcoffea/issues/99) that we already have for this same topic.

PR #213, opened 2022-03-30
B-tagging WP SFs and systematics

andrea21z:
Changes:
- I have included the WP loose in the implematation of the b-tagging SFs method (following the [recommendations](https://twiki.cern.ch/twiki/bin/viewauth/CMS/BTagSFMethods)).
- Before we had only two syst. variations (btagSFUp and btagSFDown), and now they are splited by flavor (bc or light) and by correlated/uncorrelated:
        - btagSFbc_corrUp/Down
        - btagSFlight_corrUp/Down
        - btagSFbc_uncorrUp/Down
        - btagSFlight_uncorrUp/Down

I have checked the Data/MC agreement in the ttbar CR and the syst. variations and everything looks fine!


analysis/topEFT/topeft.py
```diff
@@ -132,6 +133,7 @@ def process(self, events):
             sow_factDown       = self._samples[dataset]["nSumOfWeights_factDown"]
             sow_renormfactUp   = self._samples[dataset]["nSumOfWeights_renormfactUp"]
             sow_renormfactDown = self._samples[dataset]["nSumOfWeights_renormfactDown"]
+        '''
```
bryates, line 136: Can you remove the block comments here?

bryates (changes requested):
Thanks @andrea21z, I just had one change. I'll update the datacard to handle the correlations, but we don't need to hold off on this PR.

analysis/topEFT/topeft.py
andrea21z: yes, I  forget to uncomment it

bryates (approved):
Looks good!

andrea21z:
> @andrea21z looks like the datacard checks are failing. Do we expect these updates to affect the nominal yields?

Yes, they will change a little

bryates:
> > @andrea21z looks like the datacard checks are failing. Do we expect these updates to affect the nominal yields?
> 
> Yes, they will change a little

Ok, then can you update the reference files? Once that is done, I think this PR should be ready to merge.

bryates:
> > > > @andrea21z looks like the datacard checks are failing. Do we expect these updates to affect the nominal yields?
> > > 
> > > 
> > > Yes, they will change a little
> > 
> > 
> > Ok, then can you update the reference files? Once that is done, I think this PR should be ready to merge.
> 
> I have updated the reference json file, but I think I can't update the datacard files, those files were always updated by Kelci or you because I can't run those tests in Oviedo (or maybe other reason, I don't remember...)

Ok, I can't remember either, but I'll take care of it now.

andrea21z:
> > > > > @andrea21z looks like the datacard checks are failing. Do we expect these updates to affect the nominal yields?
> > > > 
> > > > 
> > > > Yes, they will change a little
> > > 
> > > 
> > > Ok, then can you update the reference files? Once that is done, I think this PR should be ready to merge.
> > 
> > 
> > I have updated the reference json file, but I think I can't update the datacard files, those files were always updated by Kelci or you because I can't run those tests in Oviedo (or maybe other reason, I don't remember...)
> 
> Ok, I can't remember either, but I'll take care of it now.

Thank you very much :)

kmohrman:
Thanks @andrea21z, this PR looks good to me. I just pushed two small commits:
* Update ref yields (the ones for the larger check than the CI check)
* Revert remorm/fact stuff to match the master branch (we do not include renorm and fact right now, and instead just use renormfact since we're not currently doing the envelope approach)

PR #215, opened 2022-04-04
Update to the latest version of lepton MVA

sscruz:
This PR:
+ updates the lepton MVA variable so it uses the new lepton MVA for electrons
+ updates the fakeable object definition
+ updates the to the UL fake rates (not yet in their latest version) 
+ fixes gen matching by require leptons to match the charge of the gen-matched particle
+ includes @barbaralvarez and @andrea21z latest version of the SFs
+ updates the CI to run over a skimmed sample


bryates (approved):
Thanks @sscruz, all the changes look good, and everything is passing.

topcoffea/modules/HistEFT.py
```diff
@@ -619,7 +619,7 @@ def rebin(self, old_axis, new_axis):
     """ Rebin a dense axis """
     old_axis = self.axis(old_axis)
     if isinstance(new_axis, numbers.Integral):
-        new_axis = Bin(old_axis.name, old_axis.label, old_axis.edges()[::new_axis])
+        new_axis = coffea.hist.Bin(old_axis.name, old_axis.label, old_axis.edges()[::new_axis])
```
kmohrman, line 622: This looks interesting. So you are using `Bin` from `coffea.hist` I am wondering what this changes or fixes?

kmohrman:
Thanks @sscruz, this looks good to me! I had just one quick question (above) but other than that, I'd be happy for it to be merged.

topcoffea/modules/HistEFT.py
sscruz: `Bin` was crashing because it was not imported at that point, that's why I replaced it by `coffea.hist.Bin`
I think this line was not used by other functions in the repo, is it?

kmohrman: I think we use `rebin` in the datacard maker [here](https://github.com/TopEFT/topcoffea/blob/master/analysis/topEFT/datacard_maker.py#L287), so I guess it's a bit surprising that this was not causing a crash there?

kmohrman: Interesting, I added some print statements, and it seems this line was indeed being used when we called `rebin()` in the datacard maker. It was not raising an error for me, so I am not sure why it was crashing for you. But I guess it looks like it would be using coffea's `Bin()` either way, so probably this change is fine.

PR #218, opened 2022-04-06
Update 2016APV skims and introduce missing MC background NAODv9 skim samples

Andrew42:
This PR updates the skims for data, private signal mc, and central background mc samples to properly include the lepMVA value in samples related to the 2016APV scenario. In the previous skims, the 2016APV lepMVA was not properly called by the skimmer and so was not included in those samples. The non-APV 2016, 2017, and 2018 variants were not affected by this mistake.

Also included in this PR are skims for the background samples included in PR #217.

The [make_skim_jsons.py](https://github.com/TopEFT/topcoffea/blob/master/analysis/topEFT/make_skim_jsons.py) script was also modified to handle the case where a subset of datasets are taken from a particular lobster skimming job and the remaining datasets taken from another job. An important thing to note on the implementation is that when the script is parsing the `/hadoop` directories and finds a dataset name that was already found from a different `/hadoop` directory, it _overwrites_ that entry and replaces it. This means that for this mechanism to work properly, the order of `/hadoop` directories to be iterated over (e.g. in the `.txt` file passed via the `--file` option) needs to be considered to ensure that the JSON of the corresponding datasets get updated with the correct file lists.

The python commands used to update the json files:
```
python make_skim_jsons.py --file ND_private_mc_skim_locations.txt --json-dir ../../topcoffea/json/signal_samples/private_UL
python make_skim_jsons.py --file ND_data_skim_locations.txt --json-dir ../../topcoffea/json/data_samples
```


analysis/topEFT/make_skim_jsons.py
```diff
@@ -81,17 +81,22 @@ def main():
         verbose = verbose
     )
     missing_templates = []
-    hadoop_dataset_dirs = []
+    # hadoop_dataset_dirs = []
```
bryates, line 84: Can you remove this commented line?

analysis/topEFT/make_skim_jsons.py
```diff
@@ -81,17 +81,22 @@ def main():
...
+    hadoop_dataset_dirs = {}
     for src_dir in src_dirs:
         for d in os.listdir(src_dir):
             dir_fpath = pjoin(src_dir,d)
             if not os.path.isdir(dir_fpath): continue
-            hadoop_dataset_dirs.append(dir_fpath)
+            # hadoop_dataset_dirs.append(dir_fpath)
```
bryates, line 90: Same here

analysis/topEFT/make_skim_jsons.py
```diff
@@ -81,17 +81,22 @@ def main():
...
+            if d in hadoop_dataset_dirs:
+                old = hadoop_dataset_dirs[d]
+                print(f"Replacing: {old} -> {dir_fpath}")
+            hadoop_dataset_dirs[d] = dir_fpath
     print("Skim directories:")
     for hdir in hadoop_dataset_dirs:
         print(f"\t{hdir}")
     print("Attempting to find matching json templates...")
-    for hdir in hadoop_dataset_dirs:
+    for k,hdir in hadoop_dataset_dirs.items():
```
bryates, line 99: If you're not using `k` you can use `_,hdir` or 
```suggestion
    for hdir in hadoop_dataset_dirs.values():
```

bryates (commented):
Thanks @Andrew42. I just had a couple small comments.

Andrew42:
I address the changes mentioned by @bryates and also added a checker script that can be run to make sure the lepMVA weights are present in at least one file from each of the datasets

bryates (approved):
@Andrew42 thanks for the fixes. This PR now looks good!

kmohrman:
Thanks very much @Andrew42! This looks good to me as well. I'd be happy to merge as soon as the CI finishes.

PR #219, opened 2022-04-06
Datacard maker fixes and b-tag correlations

bryates:
This PR adds the year to uncorrelated b-tag systematics, so that combine will understand they are fully uncorrelated across the years.

It also fixes an issue with the datacard maker processing the charge flips, as well as improved parallelization for the differential distributions.


andrea21z:
Looks good to me!

andrea21z (approved):

kmohrman:
Thanks @bryates. The only question I have is about the changes to the processor. I was trying to go through everywhere the the names of the systematics are changed (i.e. changed to include the year in the name) and the effects this would have in other places (e.g. in `GetBTagSF()`) and I _think_ it looks consistent. However, I could have easily missed something, so I'm wondering, just as a check, if you or @andrea21z have tried running the processor with these changes with the systematics turned on?

If not, I think we should run a quick check before merging. I don't think we'd need to process much data, just enough to make sure there are no errors due to the naming. So, I think you could probably just process the CI file, e.g. `time python run.py ../../topcoffea/json/test_samples/UL17_private_ttH_for_CI.json --do-systs`. Or maybe you've already done a check like this, but if not, I'm wondering if you could do that check just to make sure?

bryates:
> Thanks @bryates. The only question I have is about the changes to the processor. I was trying to go through everywhere the the names of the systematics are changed (i.e. changed to include the year in the name) and the effects this would have in other places (e.g. in `GetBTagSF()`) and I _think_ it looks consistent. However, I could have easily missed something, so I'm wondering, just as a check, if you or @andrea21z have tried running the processor with these changes with the systematics turned on?
> 
> If not, I think we should run a quick check before merging. I don't think we'd need to process much data, just enough to make sure there are no errors due to the naming. So, I think you could probably just process the CI file, e.g. `python run.py ../../topcoffea/json/test_samples/UL17_private_ttH_for_CI.json --do-systs`. Or maybe you've already done a check like this, but if not, I'm wondering if you could do that check just to make sure?

There was a small change to `GetBTagSF()` I forgot to include to account for the years. I know see the correct systematics, e.g.:
```python
 ('ttHJet_privateUL17',
  '2lss_m',
  'btagSFlight_2017Down',
  'isSR_2lSS'): array([0.00000000e+00, 0.00000000e+00, 0.00000000e+00, 0.00000000e+00,
        2.23470890e-04, 1.81369438e-04, 8.43592477e-05, 4.74585523e-05,
        9.90492894e-06, 1.51719933e-05]),
 ('ttHJet_privateUL17',
  '3l_CR',
  'btagSFlight_2017Down',
  'isSR_3l'): array([6.29035887e-06, 2.05587910e-05, 2.53896192e-05, 3.07534655e-05,
        2.39510411e-05, 1.28820194e-06, 2.87774758e-06, 1.56763880e-06,
        0.00000000e+00, 0.00000000e+00])
```

kmohrman:
Ok, thanks @bryates! If it now runs for you with the systematics turned on, I'd be good with it being merged now.

bryates:
> Ok, thanks @bryates! If it now runs for you with the systematics turned on, I'd be good with it being merged now.

Thanks, I'm doing one last check on the datacard maker. Once it's done, I'll merge.

PR #220, opened 2022-04-06
Fix json year names

kmohrman:
This PR updates the `year` key in the 2016 data jsons to be `2016APV` for the HIPM datasets. Similarly, it updates the `histAxisName` key to be `dataUL16APV` instead of just `dataUL16`. 

However, it does not change the actual `make_jsons.py` script, so if we were ever to remake the JSONs, we'd again end up with the wrong year names. Should we also update `make_jsons.py` as part of this PR?


bryates (approved):
Looks good, thanks @kmohrman.

PR #222, opened 2022-04-07
Adding lepton ID SFs and splitting uncertainties

andrea21z:
This PR includes:
- Lepton ID SFs:
    - For muons we take the SFs provided by the POG: [repo](https://gitlab.cern.ch/cms-muonPOG/muonefficiencies/-/tree/master/Run2/UL)
    - For electrons Bárbara computed them because we use a 'special' WP point (WPL) that has not central SFs. 

- I have splitted the lepton uncertainties in 'lepSF_muon' and 'lepSF_elec', as discussed some meetings ago.


kmohrman:
Thanks very much for the PR @andrea21z! It looks like this has an impact on the nominal yields, so I'm wondering if you could remake the reference yields and datacards? (This just involves running the `remake_ci_ref_yields.sh` and `remake_ci_ref_datacard.sh` scripts.)

bryates (approved):
Looks good, thanks @andrea21z!

PR #224, opened 2022-04-11
Change DAS query to return correct event counts

Andrew42:
This PR is to fix a bug noticed by @kmohrman, where running the `make_jsons.py` script to regenerate the JSON files for the 2016/2017/2018 data samples would incorrectly set the `nEvents`, `nGenEvents`, and `nSumOfWeights` entries to 0. After making the change, I re-ran the script a confirmed it produces identical JSON files that are currently on master.


bryates (approved):
Looks good.

kmohrman:
Looks good, thanks @Andrew42.

PR #225, opened 2022-04-12
Update of lepton fake rates

sscruz:
Update to the latest version of lepton fake rates. The difference with respect to the previous ones is that electrons have a finer granularity at low pt to model better the turn on and they are corrected by the residual non closure in simulations


analysis/topEFT/topeft.py
```diff
@@ -126,16 +126,16 @@ def process(self, events):
...
+            sow_FSRDown        = self._samples[dataset]["nSumOfWeights_FSRDown"        ]if "nSumOfWeights_FSRDown"        in self._samples[dataset] else -1
+            sow_renormUp       = self._samples[dataset]["nSumOfWeights_renormUp"       ]if "nSumOfWeights_renormUp"       in self._samples[dataset] else -1
+            sow_renormDown     = self._samples[dataset]["nSumOfWeights_renormDown"     ]if "nSumOfWeights_renormDown"     in self._samples[dataset] else -1
+            sow_factUp         = self._samples[dataset]["nSumOfWeights_factUp"         ]if "nSumOfWeights_factUp"         in self._samples[dataset] else -1
+            sow_factDown       = self._samples[dataset]["nSumOfWeights_factDown"       ]if "nSumOfWeights_factDown"       in self._samples[dataset] else -1
+            sow_renormfactUp   = self._samples[dataset]["nSumOfWeights_renormfactUp"   ]if "nSumOfWeights_renormfactUp"   in self._samples[dataset] else -1
+            sow_renormfactDown = self._samples[dataset]["nSumOfWeights_renormfactDown" ]if "nSumOfWeights_renormfactDown" in self._samples[dataset] else -1
```
bryates, line 138: Was this introduced while you were testing? Should we undo these changes, or even better, raise an exception?

bryates (commented):
Thanks @sscruz, I just had one comment.

analysis/topEFT/topeft.py
bryates: I believe all the jsons in the repo (at least for the ND T3) have the sum of weights stored. We should add them to any missing files (like the PSI version), otherwise the systematics will be wrong.

sscruz: indeed, but there are samples for which these weights are not even computed (for instance the QCD ones for the nonprompt estimation), so there's nothing to be put in the `json`. 
Of course in that case one cannot expect the systematics to work and probably we don't want to put this behavior as part of the "mainstream" code. I've rolled back the code to the previous version, but  it's reading only the weights when one wants to compute the systematics. That way i can run these samples (mostly to cross check non prompt estimations, for which I don't need these kind of systematic) and at the same time we prevent mistakes from miscalculated uncertainties

kmohrman:
Thanks for this PR @sscruz. @bryates and @sscruz I'm wondering if you know why the CI is failing? Do the reference yields need to be updated? But I'm not sure if that would be expected here.

bryates:
> Thanks for this PR @sscruz. @bryates and @sscruz I'm wondering if you know why the CI is failing? Do the reference yields need to be updated? But I'm not sure if that would be expected here.

It looks like it's failing to even make a `pkl` file [1]. I haven't had a chance to debug this.

[1] https://github.com/TopEFT/topcoffea/runs/6004998986?check_suite_focus=true#step:14:31

sscruz:
the latest change for the I/FSR SOW broke things since it's trying to access those variables somewhere later (not sure where or why since we are not using systematics). I will fix it later

bryates:
> the latest change for the I/FSR SOW broke things since it's trying to access those variables somewhere later (not sure where or why since we are not using systematics). I will fix it later

This fix actually exposes an issue. I don't think we need to bother storing those weights if we're not processing the systematics (which the CI does not). No need to hold up this PR, but I'll open an issue for it.

kmohrman:
Ok, this looks good to me. @bryates if you're also happy, should we go ahead and merge it?

bryates:
> Ok, this looks good to me. @bryates if you're also happy, should we go ahead and merge it?

Sure, I'll merge now.

PR #230, opened 2022-04-15
Implement FF up/down systematics

kmohrman:
This PR implements the up/down variations for the fake factor systematics. We had bene calculating the up/down variations, but had not actually been including them (i.e. the output data histograms only had the `nominal` category on the `systematics` axis, and were missing the `FFUp` and `FFDown` categories). 

This implementation required some changes to topeft's `self._do_systematics` logic (e.g. since previously we had been turning this flag off when `isData` was true). Even though the logic now looks a bit different, the end result should only affect the up/down variations of the data samples, and should have no impact on any other cases. The CI is passing, implying that these changes indeed do impact the MC when the systematics are turned off. Since the CI does not check other cases (e.g. data with systematics off/on or MC with systematics on) I also tried running a few test jobs on the master branch and on this branch to check these cases, and I saw consistent results. I think these tests show that the changes in logic indeed do not seem to be unintentionally impacting any other parts of the code.  

However, I do have one quick question for @sscruz. I'm wondering the up/down variations of the FF should also include the prompt subtraction? That is how I attempted to implement it, but was not one hundred percent sure if that was correct.


sscruz:
hi, 
yes, I think it'd be consistent to implement them, although I don't think that should make a big difference. I understand from this line https://github.com/TopEFT/topcoffea/blob/59748af92807d0fc1158fb967c065a34b1948061/analysis/topEFT/topeft.py#L765 that this is already implemented, isn't it?
I do have some additional comment, which is that there are three sources of uncertainty associated to the fake rate: https://github.com/TopEFT/topcoffea/blob/59748af92807d0fc1158fb967c065a34b1948061/topcoffea/modules/corrections.py#L114 (the `1/2` correspond to "up" and "down" variations of the fake rate as a function of pt (`pt1/2`) and eta (`eta1/2`)
I think it should be easy to implemented after https://github.com/TopEFT/topcoffea/blob/59748af92807d0fc1158fb967c065a34b1948061/analysis/topEFT/topeft.py#L415  
by having 
` weights_dict[ch_name].add("FFpt", np.ones_like(events.fakefactor_2l), copy.deepcopy(events.fakefactor_2lpt1/events.fakefactor_2l), copy.deepcopy(events.fakefactor_2lpt2/events.fakefactor_2l))`

And then there's an additional uncertainty in the closure, that can be implemented as a flat uncertainty depending only on the channel (see today's gdoc)

kmohrman:
Thanks @sscruz.

> I understand from this line that this is already implemented, isn't it?
Yes, I implemented it because I was guessing that was the way we'd want to do it, but was not sure. Thanks for confirming.

>  there are three sources of uncertainty associated to the fake rate
Thanks for explaining this, I am attempting to add these as you outlined, and will run some additional checks to make sure it's working the way I think it's working. I have a quick question though, what does `be` mean for the eta variations (maybe "barrel and end cap", or maybe something else?)

> And then there's an additional uncertainty in the closure, that can be implemented as a flat uncertainty depending only on the channel
I've added an issue so that we remember to put this in the data card maker.

sscruz:
> Thanks for explaining this, I am attempting to add these as you outlined, and will run some additional checks to make sure it's working the way I think it's working. I have a quick question though, what does `be` mean for the eta variations (maybe "barrel and end cap", or maybe something else?)
> 
yes, exactly, it's barrel-endcap

analysis/topEFT/topeft.py
```diff
@@ -122,10 +122,8 @@ def process(self, events):
         xsec               = self._samples[dataset]["xsec"]
         sow                = self._samples[dataset]["nSumOfWeights"]
 
-        # Turn systematics off if this is a data sample, otherwise get up down weights from input dict
-        if isData:
-            self._do_systematics = False
-        else:
```
bryates, line 128: Just making sure, this was removed because now we _do_ have data systematics (`FFUp/Down`)? Did we double check the other systemics are still only being applied to the MC?

bryates (commented):
Looks good, I just had the one question.

analysis/topEFT/topeft.py
kmohrman: Yes, the data numbers remain consistent with what we have on the master branch, so I think we should be good. I've also added an extra check [here](https://github.com/TopEFT/topcoffea/blob/745c44e455f143eb58d49eedb32a310f935577d8/analysis/topEFT/topeft.py#L780-L781) that should raise an error if any other up/down systematics are applied to data (this is in case we make some change to the systematics in the future and e.g. forget to include an `if not isData` statement, this exception should notify us that we have accidentally applied a non-data systematic fluctuation to data).

bryates: Great!

bryates (approved):

kmohrman:
Ok, I I've attempted to implement the pt and eta FF variations as discussed above. I've run some test jobs, and the numbers look reasonable to me, so hopefully it's working properly. I've also rerun the tests to compare against the master branch (to make sure none of the new changes unintentionally messed up any other parts of the code), and everything looked consistent. So I think I'd be ready to merge now, since even if there are other issues that we end up finding later on, this is at least (hopefully) a step in the right direction, and I think it'd be good to test it out to see how it affects the limits. If we end up finding more issues later on, we can always address them in a different PR.

So unless there are any objections, I'll go ahead and merge.

PR #235, opened 2022-04-19
Weights updates

kmohrman:
This PR follows up on PR #228 to duplicate the `weights` object before calling `add()` in the loop over the systematics that affect kinematics. PR 228 solved this issue for the trigger SF and the btag SF by moving these inside of the loop over lepton categories. However, I think there are two things to keep in mind about this:
1. The call to `add()` for these two systematics (trigger SF and btag SF) does not really need to be taking place inside of the loop over lepton categories, as they do not depend on the lepton category. It was just convenient to put them there since it was after we had duplicated the `weights` object.
2. The btag corr `add()` call was actually still happening before the duplication ([here](https://github.com/TopEFT/topcoffea/blob/44efcc9b0dd778d78ffac11beea00fbf079436b1/analysis/topEFT/topeft.py#L407)), so I think this still would have resulted in double counting. It would be an option to also move this inside of the loop over lepton categories, but that seems like it might become unnecessarily confusing (since this also does not depend on the lepton flavor).

To work around this issue, I've duplicated the `weights` object at the start of the loop over the systematics. This way, we can call `add()` without any issues after this.  

Unrelatedly, I've also cleaned up some whitespace in the processor.

### DIFF
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -227,7 +227,8 @@ class AnalysisProcessor(processor.ProcessorABC):
         # These weights can go outside of the outside sys loop since they do not depend on pt of mu or jets
         # We only calculate these values if not isData
         # Note: add() will generally modify up/down weights, so if these are needed for any reason after this point, we should instead pass copies to add()
-        weights_any_lep_cat = coffea.analysis_tools.Weights(len(events),storeIndividual=True)
+        # Note: Here we will to the weights object the SFs that do not depend on any of the forthcoming loops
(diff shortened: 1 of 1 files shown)


kmohrman:
Just a quick follow up on item 2 above: As pointed out by @andrea21z, this block only happens during the `nominal` iteration of the loop, so it would not technically need to happen before we duplicate the `weights` object. However, to me at least, it is still easier to follow what is happening if we duplicate the object at the beginning of the loop (for reason 1 above).

bryates (approved):

andrea21z (approved):
It looks good to me

PR #236, opened 2022-04-19
Adding Iso+IP Lepton SFs and Trigger syst. unc.

andrea21z:
This PR includes:

- Isolation and IP to Loose ID lepton SFs computed by Bárbara, as suggested by the experts. 
- A place holder of the systematic uncertainty of the Trigger SFs. I have to calculate this uncertainty, but in the meantime I think we can use 2% as a realistic approach, which is what other analyses use.


kmohrman:
Thanks @andrea21z, this PR looks good to me.

kmohrman:
@bryates, any comments on this, or should we go ahead and merge?

bryates (approved):

topcoffea/modules/corrections.py
```diff
@@ -569,8 +591,8 @@ def GetTriggerSF(year, events, lep0, lep1):
...
     ls.append(SF_ee*SF_em*SF_mm*SF_eee*SF_eem*SF_emm*SF_mmm)
-  ls[1]=np.where(ls[1]==1.0,0.0,ls[1]) # syst down
-  ls[2]=np.where(ls[2]==1.0,0.0,ls[2]) # syst up
+  ls[1]=np.where(ls[1]==1.0,0.0,ls[1]) # stat unc. down
+  ls[2]=np.where(ls[2]==1.0,0.0,ls[2]) # stat unc. up
   events['trigger_sf']=ls[0] #nominal
-  events['trigger_sfDown']=ls[0]+ls[1]
-  events['trigger_sfUp']=ls[0]+ls[2]
+  events['trigger_sfDown']=ls[0]-np.sqrt(ls[1]*ls[1]+0.02*0.02) # place holder: 2% systematic unc.
+  events['trigger_sfUp']=ls[0]+-np.sqrt(ls[2]*ls[2]+0.02*0.02) # place holder: 2% systematic unc.
```
kmohrman, line 598: @andrea21z,  I was just looking through the changes one more time before merging, and noticed something that I have a quick question about on this line. Is it actually supposed to be `+-` here, or just `+`?

andrea21z: Good catch! It's just -

kmohrman: Ok, thanks for fixing it!

Issue #231, opened 2022-04-15
Include FF rate uncertainties in data card maker


kmohrman:
As mentioned in PR #230, we need to add the relevant FF rate uncertainties to the the data card maker.

kmohrman:
These uncertainties were not actually rate uncertainties, but were implemented in the processor as shape uncertainties, handled in the datacard maker in PR #239.

PR #237, opened 2022-04-19
Play nice with fakes

bryates:
This PR fixes the `fakes` systematics (`FF`) in the datacard maker.

### DIFF
--- a/analysis/topEFT/datacard_maker.py
+++ b/analysis/topEFT/datacard_maker.py
@@ -194,7 +194,9 @@ class DatacardMaker():
                 for syst,histo in h.items():
                     if syst == 'nominal':
                         fout[name+cat] = hist.export1d(histo)
-                    elif self.do_nuisance and name not in self.syst_special and 'flips' not in name:
+                    elif self.do_nuisance and name not in self.syst_special:
+                        if 'fakes' in name and 'FF' not in syst: continue # Only processes fake factor systs for flips
+                        if 'fakes' not in name and 'FF' in syst: continue # Don't processes fake factor systs for others
                         # Special cass for systematics NOT correlated by year
                         if bool(re.search('UL\d\d', name)) and bool(re.search('20\d\d', syst)):
                             # Find systematic year
@@ -306,7 +308,7 @@ class DatacardMaker():
             if len(h_base.axes())>1:
                 fout[pname+'sm'] = export2d(h_bases)
             else:
-                if any([sig in p for sig in self.signal]):
+                if any([sig in p for sig in self.signal]) or 'fakes' in pname:
                     export1d(h_sm, pname, 'sm', fout) # Special case for SM b/c background names overlap (p not pname)
                 else:
                     export1d(h_sm, p, '_sm', fout) # Special case for SM b/c background names overlap (p not pname)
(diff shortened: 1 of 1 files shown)


analysis/topEFT/datacard_maker.py
```diff
@@ -194,7 +194,9 @@ def export1d(h, name, cat, fout):
                 for syst,histo in h.items():
                     if syst == 'nominal':
                         fout[name+cat] = hist.export1d(histo)
-                    elif self.do_nuisance and name not in self.syst_special and 'flips' not in name:
+                    elif self.do_nuisance and name not in self.syst_special:
+                        if 'fakes' in name and 'FF' not in syst: continue # Only processes fake factor systs for flips
```
kmohrman, line 198: This is a very minor question, but just to avoid confusion, should the comment say "fakes" here not "flips"?

bryates: Right, I'll fix it now.

analysis/topEFT/datacard_maker.py
```diff
@@ -194,7 +194,9 @@ def export1d(h, name, cat, fout):
...
+                        if 'fakes' not in name and 'FF' in syst: continue # Don't processes fake factor systs for others
```
kmohrman, line 199: In the pkl file there should not be FF systematics for other samples, but I guess this is a good double check of that.

analysis/topEFT/datacard_maker.py
kmohrman: Thanks

analysis/topEFT/datacard_maker.py
bryates: Ok, so the second `if` is probably redundant? Like you said, it's probably good to double check. I don't think it will cause a bottleneck.

kmohrman: Right, it's probably redundant, but I guess it does not hurt to keep it (maybe we'd even want to have a warning or something if the second if is true, just to alert us that something is wrong with the pkl file). But I'm completely fine with leaving it as is for now.

kmohrman:
Thanks @bryates, this looks good to me and when I tested it on my pkl file, it indeed included the FF systematics in the text data card and the root file. I'd be happy for the PR to be merged at this point.

However, there are other changes we will need to make to the datacard maker related to the FFs (Issue #231, adding FF rate uncertainties). If you wanted to include those in this PR, that'd be good with me. Or if you want to do that in a separate PR, that's also good with me.

bryates:
> Thanks @bryates, this looks good to me and when I tested it on my pkl file, it indeed included the FF systematics in the text data card and the root file. I'd be happy for the PR to be merged at this point.
> 
> However, there are other changes we will need to make to the datacard maker related to the FFs (Issue #231, adding FF rate uncertainties). If you wanted to include those in this PR, that'd be good with me. Or if you want to do that in a separate PR, that's also good with me.

Sure, we can include them here. If you want, just add them to:
https://github.com/TopEFT/topcoffea/blob/b1d298aeb2062058d68a6ff63c79e3690e86a775/analysis/topEFT/datacard_maker.py#L28
but if you have trouble, just send me the numbers and I can add them.

kmohrman:
Ok, thanks @bryates. I can try to add them, but there's one thing I'm confused about. The errors are different depending on the year, and I'm not sure if there's already a way to handle this in the `syst_special` dictionary?  Or if we will need to make some modifications? Or include them in a different place/way?

bryates:
> Ok, thanks @bryates. I can try to add them, but there's one thing I'm confused about. The errors are different depending on the year, and I'm not sure if there's already a way to handle this in the `syst_special` dictionary? Or if we will need to make some modifications? Or include them in a different place/way?

I think something like `'fakes': {'fakes_17: x}` might work, but I'll double check.

bryates:
> > Ok, thanks @bryates. I can try to add them, but there's one thing I'm confused about. The errors are different depending on the year, and I'm not sure if there's already a way to handle this in the `syst_special` dictionary? Or if we will need to make some modifications? Or include them in a different place/way?
> 
> I think something like `'fakes': {'fakes_17: x}` might work, but I'll double check.

Take look here:
https://github.com/TopEFT/topcoffea/blob/cd27f8c94b7a198f8e9fcbdb872f3a0255e0da86/analysis/topEFT/datacard_maker.py#L28

kmohrman:
Ok, so after further discussion with @bryates, it seems that the implementation of the per-year FF rate uncertainties will be a bit more involved than we thought (as we are not sure how to apply these uncertainties, since we the nonprompt contribution is not split by year). Also, upon rereading @sscruz's explanation of the uncertainties (in last Friday's google doc), we also realized that the uncertainties seem to need to be split by flavor. We are also not sure how this should be implemented.

So for these reasons, we decided to go ahead and merge this branch now. And then we can figure out the FF rate uncertainties in a later branch.

PR #239, opened 2022-04-20
Fix fakes in datacard maker

bryates:
This PR fixes the issue with the fakes. The recent updates was overwriting each year entry with the subsequent (e.g., if loading `nonpromptUL17` followed by `nonpromptUL18` _only_ `nonpromptUL18` would be written to the datacard). It's a bit of a hack, since it treats the fakes as special, but we can come back to this later if we need to generalize it.

### DIFF
--- a/analysis/topEFT/datacard_maker.py
+++ b/analysis/topEFT/datacard_maker.py
@@ -288,7 +288,6 @@ class DatacardMaker():
                 pyear = year[2:]
                 yproc = re.sub("UL\d\d(?:APV)?", "UL"+pyear, proc)
                 mkey = [k for k in h._sumw if proc == str(k[0]) and 'nominal' in str(k[1])]
-                #if len(mkey) == 0: continue
                 mkey = mkey[0]
                 for syst in systs:
                     if pyear != re.findall("20\d\d(?:APV)?", syst)[0][2:]: continue
@@ -298,10 +297,7 @@ class DatacardMaker():
                         if yproc == str(key[0]) or 'nominal' not in str(key[1]): continue
                         if 'nominal' in str(mkey[1]) and 'nominal' in str(key[1]): continue
                         kyear = re.findall("\d\d(?:APV)?", str(key[0]))[0]
-                        nom = nom + h._sumw[key] * self.lumi['20'+kyear]/self.lumi['20'+pyear]
-                    if mkey is None: continue
-                    if nom.shape != h._sumw[mkey].shape: continue
-                    h._sumw[mkey] = h._sumw[mkey] + nom
+                        h._sumw[mkey] += h._sumw[key] * self.lumi['20'+kyear]/self.lumi['20'+pyear]
         # Scale each plot to the SM
         processed = []
         for proc in self.samples:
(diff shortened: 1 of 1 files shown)


kmohrman:
Thanks @bryates. I'm currently testing this (as discussed on mattermost), but in the meantime I have a quick question. I'm wondering why the nonprompts have to be treated differently from the other samples to begin with (as you mentioned they are being treated as a special case)?

bryates:
> Thanks @bryates. I'm currently testing this (as discussed on mattermost), but in the meantime I have a quick question. I'm wondering why the nonprompts have to be treated differently from the other samples to begin with (as you mentioned they are being treated as a special case)?

I think it's because we're classifying the fakes as background, whereas the other systematics are applied to the signal. The way the datacard maker merges the signals by year is different. We _could_  just lump the fakes into signal, and make sure when we get to `signalcount` and `bkgcount` we give it the right number.

kmohrman:
> > Thanks @bryates. I'm currently testing this (as discussed on mattermost), but in the meantime I have a quick question. I'm wondering why the nonprompts have to be treated differently from the other samples to begin with (as you mentioned they are being treated as a special case)?
> 
> I think it's because we're classifying the fakes as background, whereas the other systematics are applied to the signal. The way the datacard maker merges the signals by year is different. We _could_ just lump the fakes into signal, and make sure when we get to `signalcount` and `bkgcount` we give it the right number.

Oh, interesting, I did not realize this. Just to make sure I'm following, you are saying the systematics are not applied to the backgrounds (just to signal)?

bryates:
> > > Thanks @bryates. I'm currently testing this (as discussed on mattermost), but in the meantime I have a quick question. I'm wondering why the nonprompts have to be treated differently from the other samples to begin with (as you mentioned they are being treated as a special case)?
> > 
> > 
> > I think it's because we're classifying the fakes as background, whereas the other systematics are applied to the signal. The way the datacard maker merges the signals by year is different. We _could_ just lump the fakes into signal, and make sure when we get to `signalcount` and `bkgcount` we give it the right number.
> 
> Oh, interesting, I did not realize this. Just to make sure I'm following, you are saying the systematics are not applied to the backgrounds (just to signal)?

Yes, which I guess we need to decide if that is correct. Here's an example from a root file I have:
```
root [3] gDirectory->ls("*sm*btagSFbc_2016Up")
TFile**         histos/ttx_multileptons-2lss_m_2b.root
 TFile*         histos/ttx_multileptons-2lss_m_2b.root
  KEY: TH1D     tHq_sm_btagSFbc_2016Up;1        Events
  KEY: TH1D     tllq_sm_btagSFbc_2016Up;1       Events
  KEY: TH1D     ttH_sm_btagSFbc_2016Up;1        Events
  KEY: TH1D     ttll_sm_btagSFbc_2016Up;1       Events
  KEY: TH1D     ttlnu_sm_btagSFbc_2016Up;1      Events
  KEY: TH1D     tttt_sm_btagSFbc_2016Up;1       Events
```

kmohrman:
Thanks @bryates! I'd be happy for this to be merged as soon as the CI finishes.

Issue #233, opened 2022-04-15
Determine if we need the up/down variations for the Rochester corrections


kmohrman:
At the Friday April 15 weekly group meeting, @rgoldouz mentioned that it is possible that it may not be recommended to include the up/down variations for the Rochester corrections. It would be good if someone could check into this. @andrea21z, is this something you'd be able to do?

Andrew42:
So I was messing around with (crudely) benchmarking the `topeft` processor and found some interesting results relating to the Rochester corrections. It appears as though a large fraction of the _entire_ processing time is concentrated in just two lines of code that are a part of the Rochester correction computation.

Some notes and caveats to keep in mind regarding these timings:

- Dataset: ttllJet (skimmed)
- Num. Files: 8
- Num. Chunks: 8
- Num. Workers: 8

Timings computed inside the processor were summed up and divided by the number of chunks. The time intervals were determined using `time.time()`. Rows marked with an asterisk (`*`) denote timings the span multiple lines of code and can be a superset of some of the other rows in the table. The percentages given next to each timing is the fraction of the "Total" row. All tests were done using the `futures` executor and ran from `earth`.

| Chunksize | 500 | 5k  | 50k | 100k |
| --- | --: | --: | --: | --: |
| [roch_calcs_init*](https://github.com/TopEFT/topcoffea/blob/master/topcoffea/modules/corrections.py#L477-L481) |  0.04 s   (0%) |  0.12 s   (0%) |   0.76 s  (<1%) |   1.64 s   (1%) |
| [mc_kspread](https://github.com/TopEFT/topcoffea/blob/master/topcoffea/modules/corrections.py#L483)            |  0.11 s   (0%) |  0.13 s   (0%) |   0.32 s   (0%) |   0.56 s   (0%) |
| [mc_ksmear](https://github.com/TopEFT/topcoffea/blob/master/topcoffea/modules/corrections.py#L484)             |  0.18 s   (0%) |  0.19 s   (0%) |   0.28 s   (0%) |   0.43 s   (0%) |
| [errspread](https://github.com/TopEFT/topcoffea/blob/master/topcoffea/modules/corrections.py#L485)             |  9.19 s  (15%) |  9.67 s  (14%) |  14.21 s  (13%) |  19.08 s  (13%) |
[...]

PR #241, opened 2022-04-22
Remove Rochester corrections systematic

Andrew42:
As per issue #233, it was decided that we can drop the up/down fluctuations for the Rochester corrections. This PR removes the steps involved with calculating those fluctuations and also moves the muon/tau/loose/FO selections back outside of the `syst_var` loop as these objects no longer depend on anything from that loop and can just be calculated a single time, instead of once per systematic fluctuation.

I've also shuffled around where the Tau selection occurs to move it closer to where the Electron and Muon selection happens as I believe it only depends on the loose selection criteria and nothing else. However, I would like to make sure someone double checks that this is the case.

### DIFF
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -211,7 +211,7 @@ class AnalysisProcessor(processor.ProcessorABC):
 
         ################### Muon selection ####################
 
-        mu["pt"] = ApplyRochesterCorrections(year, mu, isData, var='nominal') # Need to apply corrections before doing muon selection
+        mu["pt"] = ApplyRochesterCorrections(year, mu, isData) # Need to apply corrections before doing muon selection
         mu["isPres"] = isPresMuon(mu.dxy, mu.dz, mu.sip3d, mu.eta, mu.pt, mu.miniPFRelIso_all)
         mu["isLooseM"] = isLooseMuon(mu.miniPFRelIso_all,mu.sip3d,mu.looseId)
         mu["isFO"] = isFOMuon(mu.pt, mu.conept, mu.btagDeepFlavB, mu.mvaTTHUL, mu.jetRelIso, year)
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -211,9 +211,7 @@ class AnalysisProcessor(processor.ProcessorABC):
 
         ################### Muon selection ####################
 
-        # Rochester corrections
-        mu["pt"] = ApplyRochesterCorrections(year, mu, isData, var='nominal')
-
+        mu["pt"] = ApplyRochesterCorrections(year, mu, isData, var='nominal') # Need to apply corrections before doing muon selection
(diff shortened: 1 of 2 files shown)


bryates (changes requested):
Looks like this was made before the latest updates. @andrea21z can you please pull the updated master changes and merge them in here?

bryates:
> Looks like this was made before the latest updates. @andrea21z can you please pull the updated master changes and merge them in here?

Sorry, maybe I was confusing the tau selection movement with the new FF code. Ignore my request.

analysis/topEFT/topeft.py
```diff
@@ -261,61 +302,12 @@ def process(self, events):
         else: syst_var_list = ['nominal']
 
         # Loop over the list of systematic variations we've constructed
-        mu["pt_raw"]=mu.pt
```
bryates, line 264: Was the raw pT simply not used anywhere?

Andrew42: My understanding was that the line in question was used as a placeholder to store the "original" muon pt values so that they could be used to reset `mu["pt"]` on each pass through the `for` loop.

Since we're no longer have to re-compute the corrected muon pt for the up/down fluctuations on the Rochester corrections, I don't think we need to cache this variable anymore.

kmohrman: I agree, I think that copy would no longer be needed.

topcoffea/modules/corrections.py
```diff
@@ -513,25 +513,19 @@ def ApplyRochesterCorrections(year, mu, is_data, var='nominal'):
...
-        errors[~hasgen_flat] = np.array(ak.flatten(errsmear))
         corrections = ak.unflatten(corrections, ak.num(mu.pt, axis=1))
-        errors = ak.unflatten(errors, ak.num(mu.pt, axis=1))
     else:
         corrections = rochester.kScaleDT(mu.charge, mu.pt, mu.eta, mu.phi)
-        errors = rochester.kScaleDTerror(mu.charge, mu.pt, mu.eta, mu.phi)
-    if var=='nominal': return(mu.pt * corrections) #nominal
-    elif var=='up': return(mu.pt * corrections + mu.pt * errors) #up 
-    elif var=='down': return(mu.pt * corrections - mu.pt * errors) #down
+    if var == 'nominal':
```
kmohrman, line 525: Since `var` seems to be no longer used, could you remove it from the function arguments?

bryates (approved):

topcoffea/modules/corrections.py
Andrew42: Removed it!

kmohrman:
Thanks for this PR @Andrew42, it looks good to me. I've run a quick test to compare the (non Rochester) systematics here and on the master branch (just to make sure there were not any obvious unexpected changes), and everything I checked looked consistent. 

I'd be happy to merge the PR now.

PR #244, opened 2022-04-24
Always enforce lep flavor requirements for CRs

kmohrman:
This PR updates the deffiniton of the charge flip CR and the ttbar CR to include the lepton flavor requirements. Previously, these requirements had been enforced by the `lep_flav_lst` key of the CR selection dictionary. However, this requirement is only applied when running with the `--split-lep-flavor` option, so when making CR plots we had to always remember to run with that option. This has been ok up till now (except that it's a bit annoying to remember to do). However, this option requires more memory (since we it means we keep track of many more histograms), and now that we want to start looking at the systematics in the CRs, I think we no longer want to be tied to the `--split-lep-flavor` option for the CRs (since running with systematics turned on also requires a lot more memory). 

So in summary, this PR allows us to make our CR plots without `--split-lep-flavor` turned on.


bryates (approved):
Looks good.

PR #246, opened 2022-04-26
add file:// in read_cfg_file for root files in local filesystems

btovar:
E.g:

```
file:///scratch365/myuser/myrootfiles/
some.json
```

### DIFF
--- a/topcoffea/modules/utils.py
+++ b/topcoffea/modules/utils.py
@@ -113,6 +113,8 @@ def read_cfg_file(fpath,cfg={},max_files=0):
                 # Note: This implicitly assumes that a redirector line will appear before any json
                 #   paths in the cfg file
                 xrd_src = l
+            if l.startswith("file://"):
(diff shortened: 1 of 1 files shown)


bryates (approved):

topcoffea/modules/utils.py
```diff
@@ -113,6 +113,8 @@ def read_cfg_file(fpath,cfg={},max_files=0):
                 # Note: This implicitly assumes that a redirector line will appear before any json
                 #   paths in the cfg file
                 xrd_src = l
+            if l.startswith("file://"):
```
kmohrman, line 116: Hmm, I might just be thinking about this wrong, but should this be an `elif`?

btovar: Ah, yes you are correct. I think logically it does not matter much as the statement can't be true in the two tests, but that was just luck.

PR #247, opened 2022-04-26
To Runner() interface

btovar:
For a while now the `run_uproot_job` function in coffea has been deprecated. This pr updates the calls to `run_uproot_job` to the new `Runner()` interface. Some of the advantages are that the arguments for the executor and the actual run are separated, and that that the arguments to the executor are now checked by python.

### DIFF
--- a/analysis/btagMCeff/run.py
+++ b/analysis/btagMCeff/run.py
@@ -64,9 +64,11 @@ if __name__ == '__main__':
 
   processor_instance = btagMCeff.AnalysisProcessor(samplesdict)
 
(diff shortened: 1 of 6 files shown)


bryates (approved):

kmohrman:
Thank you very much @btovar for this PR! Sorry for being slow to review the it. Anyway, I've looked through the changes and also checked it out and run a couple quick tests, and it all looks good to me. It also seems to be passing all of the relevant CI, so that's good too. I'd be happy to merge it at this point.

PR #249, opened 2022-04-29
Uncorrelated uncertainties

andrea21z:
Updating uncorrealted uncertainties with _{year} in the name:
- Trigger
- JER

### DIFF
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -255,10 +255,10 @@ class AnalysisProcessor(processor.ProcessorABC):
 
         # Define the lists of systematics we include
         obj_correction_syst_lst = [
-            'JERUp','JERDown','JESUp','JESDown' # Systs that affect the kinematics of objects
+            f'JER_{year}Up',f'JER_{year}Down','JESUp','JESDown' # Systs that affect the kinematics of objects
         ]
         wgt_correction_syst_lst = [
-            "lepSF_muonUp","lepSF_muonDown","lepSF_elecUp","lepSF_elecDown",f"btagSFbc_{year}Up",f"btagSFbc_{year}Down","btagSFbc_corrUp","btagSFbc_corrDown",f"btagSFlight_{year}Up",f"btagSFlight_{year}Down","btagSFlight_corrUp","btagSFlight_corrDown","PUUp","PUDown","PreFiringUp","PreFiringDown","triggerSFUp","triggerSFDown", # Exp systs
+            "lepSF_muonUp","lepSF_muonDown","lepSF_elecUp","lepSF_elecDown",f"btagSFbc_{year}Up",f"btagSFbc_{year}Down","btagSFbc_corrUp","btagSFbc_corrDown",f"btagSFlight_{year}Up",f"btagSFlight_{year}Down","btagSFlight_corrUp","btagSFlight_corrDown","PUUp","PUDown","PreFiringUp","PreFiringDown",f"triggerSF_{year}Up",f"triggerSF_{year}Down", # Exp systs
             "FSRUp","FSRDown","ISRUp","ISRDown","renormfactUp","renormfactDown", # Theory systs (do not include "renormUp","renormDown","factUp","factDown" for now since not using envelope)
         ]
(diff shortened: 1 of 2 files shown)


kmohrman:
Thanks very much for the PR @andrea21z! @bryates, will this require any changes in the datacard maker to handle these uncorrelated systematics? Or, since they're in the same format as the other uncorrelated shape systematics, will they be handled correctly automatically? Should we do a check before merging to make sure the templates look as we would expect?

bryates:
> Thanks very much for the PR @andrea21z! @bryates, will this require any changes in the datacard maker to handle these uncorrelated systematics? Or, since they're in the same format as the other uncorrelated shape systematics, will they be handled correctly automatically? Should we do a check before merging to make sure the templates look as we would expect?

I believe it will be handled automatically.

kmohrman:
> > Thanks very much for the PR @andrea21z! @bryates, will this require any changes in the datacard maker to handle these uncorrelated systematics? Or, since they're in the same format as the other uncorrelated shape systematics, will they be handled correctly automatically? Should we do a check before merging to make sure the templates look as we would expect?
> 
> I believe it will be handled automatically.

Ok, sounds good. So then I think just to double check this, we'd want to do the following:
* Make a pkl file that include the newly uncorrelated uncertainties (could run with all systematics included, or could comment out most of the others to keep the pkl file smaller). I think we'd only need one sample, but I think we'd need all 4 years of it. 
* Run the pkl file through the datacard maker
* Run the `parse_datacard_templtes.py` script on the output templates, and check that the nominal and up and down uncertainties look as expected.

@bryates or @andrea21z, could you run this check if you have time? If you don't have time, I can try to do it at some point.

kmohrman:
> > > Thanks very much for the PR @andrea21z! @bryates, will this require any changes in the datacard maker to handle these uncorrelated systematics? Or, since they're in the same format as the other uncorrelated shape systematics, will they be handled correctly automatically? Should we do a check before merging to make sure the templates look as we would expect?
> > 
> > 
> > I believe it will be handled automatically.
> 
> Ok, sounds good. So then I think just to double check this, we'd want to do the following:
> 
> * Make a pkl file that include the newly uncorrelated uncertainties (could run with all systematics included, or could comment out most of the others to keep the pkl file smaller). I think we'd only need one sample, but I think we'd need all 4 years of it.
> * Run the pkl file through the datacard maker
> * Run the `parse_datacard_templtes.py` script on the output templates, and check that the nominal and up and down uncertainties look as expected.
> 
> @bryates or @andrea21z, could you run this check if you have time? If you don't have time, I can try to do it at some point.

@bryates or @andrea21z, just wondering if you've had a chance to look into this check?

bryates:
> > > > Thanks very much for the PR @andrea21z! @bryates, will this require any changes in the datacard maker to handle these uncorrelated systematics? Or, since they're in the same format as the other uncorrelated shape systematics, will they be handled correctly automatically? Should we do a check before merging to make sure the templates look as we would expect?
> > > 
> > > 
> > > I believe it will be handled automatically.
> > 
> > 
> > Ok, sounds good. So then I think just to double check this, we'd want to do the following:
> > 
> > * Make a pkl file that include the newly uncorrelated uncertainties (could run with all systematics included, or could comment out most of the others to keep the pkl file smaller). I think we'd only need one sample, but I think we'd need all 4 years of it.
> > * Run the pkl file through the datacard maker
> > * Run the `parse_datacard_templtes.py` script on the output templates, and check that the nominal and up and down uncertainties look as expected.
> > 
> > @bryates or @andrea21z, could you run this check if you have time? If you don't have time, I can try to do it at some point.
> 
> @bryates or @andrea21z, just wondering if you've had a chance to look into this check?

Sorry, I guess I missed the last comment where you asked if I could take a look. I'll make a `pkl` file now, and try to run the rest.

bryates:
I've made a full set of plots [here](https://www.crc.nd.edu/~byates2/coffea/ttlnu/) showing the shifts in the uncorrelated trigger systematics are very small. We should be good to merge now.

kmohrman:
Great, thanks for the check @bryates, I'll go ahead and merge now.

Issue #252, opened 2022-05-03
Update placeholder ZGToLLG xsec


kmohrman:
The xsec for ZGToLLG in the `xsec.cfg` needs to be updated (right now just using the `xsec.cfg`'s `ZG` number, whose origin is also unknown).

Andrew42:
See PR #255

PR #253, opened 2022-05-03
Adding jsons for ZG samples

kmohrman:
Note that the xsec being used for ZGToLLG is currently a placeholder and will need to be updated (Issue #252).


bryates (approved):
Thanks for adding these.

PR #257, opened 2022-05-06
Miscellaneous updates

kmohrman:
This PR includes many miscellaneous updates and bug fixes. To name just a few of the larger ones:
* Implements the ad-hoc solution for the selected WC issue in the datacard maker/model (by hard coding a check to make sure `ctlTi` does not enter for `tttt`)
* Bug fix for the way we construct the list of `conversions` samples in the processor
* Bug fix for njets criteria for 3l CR
* Updates the WZ xsec to inlaced the NNLO k factor 
* Updates to binning in the processor and datacard maker
* Update official set of histograms in our "analysis" list setting for the processor's `--hist-list option` (since we've decided on `ptz` and `lj0pt` as our official differential variables)


bryates (approved):
Looks good, thanks for taking care of all of these!

topcoffea/cfg/mc_background_samples.cfg
```diff
@@ -3,17 +3,17 @@
 # Central UL16APV background samples
 #root://ndcms.crc.nd.edu/
 root://deepthought.crc.nd.edu/
-../../topcoffea/json/background_samples/central_UL/UL16APV_ZGToLLG.json
-#../../topcoffea/json/background_samples/central_UL/UL16APV_TTTo2L2Nu.json
-#../../topcoffea/json/background_samples/central_UL/UL16APV_TTToSemiLeptonic.json
+#../../topcoffea/json/background_samples/central_UL/UL16APV_ZGToLLG.json
```
bryates, line 6: Are the `ZG` conversions just for the CR plots, or do we need to uncomment them.

kmohrman: The ZG conversions eventually may be used for the SR as well, but first we'll probably want to check how large the effect is in of the SR channels. So, for this run, I was not going to include them, so left it commented for now.

bryates: Sounds good.

PR #258, opened 2022-05-07
Update ignore list in datacard maker

kmohrman:
This very minor PR updates the list of ignored histos in the datacard maker to include the new ttbar powheg samples (since they have a different name than the old MG ones had, they were not being ignored). 

Mostly unrelatedly, this PR also includes a minor update to the `parse_datacard_templtes.py` script, to make it easier to dump the names of _all_ of the histos in the template root files (e.g. for checking that that all of the samples are being ignored).


bryates (approved):
Looks good.

PR #261, opened 2022-05-10
Small fixes

bryates:
This PR fixes a few things:
1. A better handling of the data (skipping when needed) in the datacard maker
2. Added `--asimov` and `--unblind` flags (currently still uses Asimov data if neither flag is set) **DO NOT fit with the unblind flag until we're unblinded**
3. Added the `WZ` powheg samples to `make_json.py`
4. Fixed a typo in `missing_parton.py`


kmohrman:
@bryates am I understanding correctly that you've replaced the old WZ samples with the new ones (and not just added the new ones)? If so, then I think we'll need to run a check of the CRs before we merge this PR, to make sure we're not surprised by any unexpected differences in the future.

bryates:
> @bryates am I understanding correctly that you've replaced the old WZ samples with the new ones (and not just added the new ones)? If so, then I think we'll need to run a check of the CRs before we merge this PR, to make sure we're not surprised by any unexpected differences in the future.

I had replaced them to keep the list simpler, but maybe we should keep both, unless until we've checked the CR plots.

bryates:
> > @bryates am I understanding correctly that you've replaced the old WZ samples with the new ones (and not just added the new ones)? If so, then I think we'll need to run a check of the CRs before we merge this PR, to make sure we're not surprised by any unexpected differences in the future.
> 
> I had replaced them to keep the list simpler, but maybe we should keep both, unless until we've checked the CR plots.

@kmohrman I've updated it so now the original `WZ` samples are there, and then new ones have `powheg` in the dictionary key.

kmohrman:
@bryates since these samples are added by @abasnet97's branch, why do we need to include them here?

bryates:
> @bryates since these samples are added by @abasnet97's branch, why do we need to include them here?

I guess we don't. Would you like me to remove them from this PR?

kmohrman:
Thanks for adding tZq @bryates! I'm happy for it to be merged once the CI finishes

Issue #266, opened 2022-05-13
Un pin pip version


kmohrman:
Once the pip/poncho incompatibility (documented [here](https://github.com/cooperative-computing-lab/cctools/issues/2856)) is resolved, we should un pin the version.

kmohrman:
Resolved by PR #273.

PR #267, opened 2022-05-13
Pin pip version

kmohrman:
This PR temporarily pins the pip version as suggested by @klannon. This temporary fix (which is applied in `remote_environment.py`) is to work around the current pip/poncho incompatibility documented [here](https://github.com/cooperative-computing-lab/cctools/issues/2856). Once that issue is resolved, we should un pin the version as mentioned in #266.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,4 +9,3 @@ dependencies:
   - dill
   - xrootd
   - git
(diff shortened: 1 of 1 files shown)


bryates (approved):
Thanks @kmohrman!

PR #268, opened 2022-05-15
Jet scale

bryates:
This PR adds the `njets` uncertainty for the `WZ` sample. Currently, all distributions are treated with a shape uncertainty, but we should be able to switch the differential distributions to a rate uncertainty.


klannon:
I'm confused by the statement that this is a "shape" uncertainty.  We need this to be a **rate** uncertainty applied to each jet category.

bryates:
> I'm confused by the statement that this is a "shape" uncertainty. We need this to be a **rate** uncertainty applied to each jet category.

It is, it's just a matter of how it fits in the datacard. For the `njets` distribution we have to provide a histogram with each bin moved, since combine only allows a single number (or pairs) for a rate uncertainty.

bryates:
Ok, this branch is now fully working. The `njets` distributions get new up/down histograms, and I've confirmed their bin contents are the same multiples (or 1/factor for the down) as the numbers derived from @kmohrman's data-to-MC ratios. For the differential distributions, there is a single rate parameter added to the datacard corresponding to the integral of the `WZ` histogram multiplied by the data-to-MC ratio.
I've also run checks against the master branch, and nothing else changes.

kmohrman:
Thanks for the update @bryates! I am wondering about a possible edge case... what about if we want to use njets for some categories, and a differential variable for other categories. Is the name of the uncertainty in the datacard the same? Is it an issue that the uncertainty would be a shape uncertainty in some categories and a rate uncertainty in others?

bryates:
> Thanks for the update @bryates! I am wondering about a possible edge case... what about if we want to use njets for some categories, and a differential variable for other categories. Is the name of the uncertainty in the datacard the same? Is it an issue that the uncertainty would be a shape uncertainty in some categories and a rate uncertainty in others?

That we would have to check in combine. I'm not sure how it handles a rate and shape with the same name, but I could give it different names.

klannon:
If the uncertainty has different names, then it would be uncorrelated, so you can't solve this by giving it different names.  I think it you have a mixed rate + shape uncertainty then either it would just work in combine that way (this would need to be checked) or you'd have to implement it as a shape uncertainty everywhere.  (In the regular bins, the "shape" uncertainty would just change the rate but not the shape.)  Was that the original implementation?  I'm sorry if my objection lead you down a wrong path.

bryates:
> > Thanks for the update @bryates! I am wondering about a possible edge case... what about if we want to use njets for some categories, and a differential variable for other categories. Is the name of the uncertainty in the datacard the same? Is it an issue that the uncertainty would be a shape uncertainty in some categories and a rate uncertainty in others?
> 
> That we would have to check in combine. I'm not sure how it handles a rate and shape with the same name, but I could give it different names.

Ah I already did sort of by accident. The `njets` version is called `jet_scale` (it adds `Diboson_sm` to look it up in the root file) in the datacard and the differential one is `Diboson_flat`. This will mean 2 nuisance parameters if we fit `njets` and differential distributions together.

bryates:
> If the uncertainty has different names, then it would be uncorrelated, so you can't solve this by giving it different names. I think it you have a mixed rate + shape uncertainty then either it would just work in combine that way (this would need to be checked) or you'd have to implement it as a shape uncertainty everywhere. (In the regular bins, the "shape" uncertainty would just change the rate but not the shape.) Was that the original implementation? I'm sorry if my objection lead you down a wrong path.

That's a good point. I'll make sure the names match and we can see how combine handles it.

bryates:
> > If the uncertainty has different names, then it would be uncorrelated, so you can't solve this by giving it different names. I think it you have a mixed rate + shape uncertainty then either it would just work in combine that way (this would need to be checked) or you'd have to implement it as a shape uncertainty everywhere. (In the regular bins, the "shape" uncertainty would just change the rate but not the shape.) Was that the original implementation? I'm sorry if my objection lead you down a wrong path.
> 
> That's a good point. I'll make sure the names match and we can see how combine handles it.

Ok, I had to make a change to all the rate names. They no longer have `_flat` in the name, which was just an internal convention I was using. If we have anything that had that name hard coded, we'll need to remove it. Otherwise, things should be fine.

kmohrman:
@bryates, does this mean you are going with the mixed rate + shape uncertainty approach? I think @klannon mentioned we'll need to check how combine handles this. I don't think this is something we can check as part of Iteration 2, I think it would need to be checked and tested independently. Is that something you'd be able to do?

bryates:
> @bryates, does this mean you are going with the mixed rate + shape uncertainty approach? I think @klannon mentioned we'll need to check how combine handles this. I don't think this is something we can check as part of Iteration 2, I think it would need to be checked and tested independently. Is that something you'd be able to do?

Sure, I can run a check.

bryates:
> > @bryates, does this mean you are going with the mixed rate + shape uncertainty approach? I think @klannon mentioned we'll need to check how combine handles this. I don't think this is something we can check as part of Iteration 2, I think it would need to be checked and tested independently. Is that something you'd be able to do?
> 
> Sure, I can run a check.

I did a very quick test and it seems to make the workspace and fit just fine. The combined card does have `shape?` in it, with the correct numbers. I'll see if `shape?` means it converted the rates to shapes for us.

bryates:
> > > @bryates, does this mean you are going with the mixed rate + shape uncertainty approach? I think @klannon mentioned we'll need to check how combine handles this. I don't think this is something we can check as part of Iteration 2, I think it would need to be checked and tested independently. Is that something you'd be able to do?
> > 
> > 
> > Sure, I can run a check.
> 
> I did a very quick test and it seems to make the workspace and fit just fine. The combined card does have `shape?` in it, with the correct numbers. I'll see if `shape?` means it converted the rates to shapes for us.

The combine documentation says it's fine
"If you have a nuisance parameter which has shape effects (using shape) and rate effects (using lnN) you should use a single line for the systemstic uncertainty with shape?. This will tell combine to fist look for Up/Down systematic templates for that process and if it doesnt find them, it will interpret the number that you put for the process as a lnN instead."

bryates:
> Thanks for the checks @bryates. I'm wondering if you could say a bit more about how you checked this? Does it have `shape?` in all cases? Or just when you mix njets and differential cards?

The documentation says it will use the correct type: so if it can find histograms it will use shape for that channel, otherwise it will use rate. 
My test just verified that I could make the workspace and run the fits with no issues.

bryates:
I've fixed the `jet_scale` uncertainties. For `njets` it now adds the nominal diboson values for all other samples (besides `WZ`) to the `jet_scale` uncertainty. For the differential distributions, the rate uncertainty is now computed as `(diboson + wz_unc) / diboson`. I've printed out values by hand for both a `WZ` only `pkl` file and a full diboson `pkl` file and both work as expected.

The `njets` plots are [here](https://www.crc.nd.edu/~byates2/coffea/WZ/)

kmohrman:
Thanks @bryates. I am a still a bit confused about a couple things though. It seems the jet scale uncertainty is now being included for tribosons? Is that expected? (It also seems to be 0 in those cases). Also, I'm a bit worried about plot [like this](https://www.crc.nd.edu/~byates2/coffea/WZ/3l2b_p/Diboson_sm_jet_scale.png). It looks like "Up" is _much_ larger than nominal.

bryates:
> Thanks @bryates. I am a still a bit confused about a couple things though. It seems the jet scale uncertainty is now being included for tribosons? Is that expected? (It also seems to be 0 in those cases). Also, I'm a bit worried about plot [like this](https://www.crc.nd.edu/~byates2/coffea/WZ/3l2b_p/Diboson_sm_jet_scale.png). It looks like "Up" is _much_ larger than nominal.

That was a mistake that is now fixed, I just need to remake the plots.

bryates:
> > Thanks @bryates. I am a still a bit confused about a couple things though. It seems the jet scale uncertainty is now being included for tribosons? Is that expected? (It also seems to be 0 in those cases). Also, I'm a bit worried about plot [like this](https://www.crc.nd.edu/~byates2/coffea/WZ/3l2b_p/Diboson_sm_jet_scale.png). It looks like "Up" is _much_ larger than nominal.
> 
> That was a mistake that is now fixed, I just need to remake the plots.

I agree the plot you linked looks worrisome. I believe it's just down to how the bins get scaled, but let me print out the numbers on that one to double check.

bryates:
> > > Thanks @bryates. I am a still a bit confused about a couple things though. It seems the jet scale uncertainty is now being included for tribosons? Is that expected? (It also seems to be 0 in those cases). Also, I'm a bit worried about plot [like this](https://www.crc.nd.edu/~byates2/coffea/WZ/3l2b_p/Diboson_sm_jet_scale.png). It looks like "Up" is _much_ larger than nominal.
> > 
> > 
> > That was a mistake that is now fixed, I just need to remake the plots.
> 
> I agree the plot you linked looks worrisome. I believe it's just down to how the bins get scaled, but let me print out the numbers on that one to double check.

Ok, there was a problem with unintended histograms being added. I've fixed the loop, and when I printed out the names of the histograms it only showed the other nominal diboson samples. The plot pointed out before is [here](https://www.crc.nd.edu/~byates2/coffea/WZ/3l2b_p/Diboson_sm_jet_scale.png) and looks much more reasonable.

kmohrman:
Thanks @bryates. Looking at the plot you linked, I'm wondering about the 2j bin. It looks like the up is somewhere around 2.6 times the nominal value. Since diboson is mostly WZ, I would have thought that in most cases the uncertainties would match pretty closely to the scaling factors we took from the CR plots (which would be 1.19 for this bin, right?). Sorry if I'm just thinking about this the wrong way, but I'm having a hard time interpreting the plot.

bryates:
> Thanks @bryates. Looking at the plot you linked, I'm wondering about the 2j bin. It looks like the up is somewhere around 2.6 times the nominal value. Since diboson is mostly WZ, I would have thought that in most cases the uncertainties would match pretty closely to the scaling factors we took from the CR plots (which would be 1.19 for this bin, right?). Sorry if I'm just thinking about this the wrong way, but I'm having a hard time interpreting the plot.

Looks like I had a nested loop when it should not have been. Fixing it now gives
```python
bin 1   9.17416 1.7309  0.507572                                             bin 2   3.46785 2.54867 0.357694                                             bin 3   1.12546 2.31756 0.401156                                             bin 4   0.19432 2.28695 0.391836
```
for the `2lss_p` plot, which is close to the dominant `WZ` ratio as mentioned. I'll update the rest of the plots when they finish.

bryates:
> > Thanks @bryates. Looking at the plot you linked, I'm wondering about the 2j bin. It looks like the up is somewhere around 2.6 times the nominal value. Since diboson is mostly WZ, I would have thought that in most cases the uncertainties would match pretty closely to the scaling factors we took from the CR plots (which would be 1.19 for this bin, right?). Sorry if I'm just thinking about this the wrong way, but I'm having a hard time interpreting the plot.
> 
> Looks like I had a nested loop when it should not have been. Fixing it now gives
> 
> ```python
> bin 1   9.17416 1.7309  0.507572                                             bin 2   3.46785 2.54867 0.357694                                             bin 3   1.12546 2.31756 0.401156                                             bin 4   0.19432 2.28695 0.391836
> ```
> 
> for the `2lss_p` plot, which is close to the dominant `WZ` ratio as mentioned. I'll update the rest of the plots when they finish.

The updated plots are [here](https://www.crc.nd.edu/~byates2/coffea/WZ/).

kmohrman:
@bryates it looks like there's still [at least one case](https://www.crc.nd.edu/~byates2/coffea/WZ/2lss_4t_m_2b/Diboson_sm_jet_scale.png/) where the up and down are both on the same side of nominal.

kmohrman:
The [4l plot](https://www.crc.nd.edu/~byates2/coffea/WZ/4l_2b/Diboson_sm_jet_scale.png) also looks a bit concerning. Maybe I am looking at an outdated version of the plots?

bryates:
> @bryates it looks like there's still [at least one case](https://www.crc.nd.edu/~byates2/coffea/WZ/2lss_4t_m_2b/Diboson_sm_jet_scale.png/) where the up and down are both on the same side of nominal.

I do see some negative bins in that channel. I'll make a pkl file with the powheg samples and see if this improve.

bryates:
> > @bryates it looks like there's still [at least one case](https://www.crc.nd.edu/~byates2/coffea/WZ/2lss_4t_m_2b/Diboson_sm_jet_scale.png/) where the up and down are both on the same side of nominal.
> 
> I do see some negative bins in that channel. I'll make a pkl file with the powheg samples and see if this improve.

So with the powheg samples I get [this](https://www.crc.nd.edu/~byates2/coffea/WZ/). There are a few cases where the `jet_scale` is identical to the nominal. In these cases, there were 0 `WZ` powheg events, so only the nominal other diboson events are added in, resulting in equal contents.

PR #269, opened 2022-05-16
Add ext backgournd samples

kmohrman:
This PR adds some extra stats for some of the background samples. Note that it will need to be skimmed and tested before we can merge.


bryates (approved):
Looks good.

Andrew42:
I've pushed the updated skim json files to use the newly skimmed versions of the ext samples.

The command used to (re-)generate the json files was:
```
```

kmohrman:
As discussed [on mm](https://mattermost.web.cern.ch/topeft-runii/pl/agj4zs743bfkjrgmjroi33nfgc), the extra stats samples have resolved the negative yields issues for tribosons. However, unfortunately, we still have histograms with negative yields for TTG.

Unless there are objections, I think it would make sense to merge this PR at this point.

PR #272, opened 2022-05-18
Missing parton

bryates:
This PR adds the missing parton uncertainty. It is derived from `tllq` `njets` and applied to both `tllq` and `tHq` for _all_ distributions.


bryates:
This PR should be good to go. All the plots are [here](https://www.crc.nd.edu/~byates2/coffea/missing_parton/datacard/) and look consistent. Any time the bin content goes negative, I just set it to zero.

PR #274, opened 2022-05-22
Jsons for LO ttgamma samples

kmohrman:
This PR adds JSONs for the new LO ttGamma samples we want to try to use for modeling the conversions background (we are hoping that the extra statistics and lack of negative weights in these LO samples will allow us to model this better than we were able to with the NLO sample we had been trying to use previously).

Two important things to note about this PR:
* This PR should not be merged until the new samples have been skimmed
* Currently, I just put in placeholder numbers for the cross sections for these samples, but we will need to find the appropriate cross section numbers before we merge.


topcoffea/cfg/xsec.cfg
```diff
@@ -29,6 +29,10 @@ WG                      : 178.8
 TTG                     : 3.697
 TTGJets                 : 3.697
 
+# PLACEHOLDERS!!! NEED TO UPDATE!!!
+TTGamma_SingleLept : 3.697
```
bryates, line 33: We should probably just scale these by similar (but slightly more accurate) calculations to what @klannon did to estimate the number of semileptonic and dileptonic events in our old `TTG` sample, unless we can find explicit ones for these.

kmohrman: I've updated the numbers as we discussed on MM.

bryates (approved):
Looks good.

Andrew42:
The skims just finished and I've created the skim versions of the corresponding JSON files. The skim locations file has also been updated.
```
```

topcoffea/json/rate_systs.json
```diff
@@ -35,7 +34,7 @@
         "tllq":     {"pdf_scale": "qq", "qcd_scale": "V"   },
         "Diboson":  {"pdf_scale": "qq", "qcd_scale": "VV"  },
         "Triboson": {"pdf_scale": "qq", "qcd_scale": "VVV" },
-        "convs":    {"pdf_scale": "gg", "qcd_scale": "ttG" }
+        "convs":    {"pdf_scale": "gg", }
```
bryates, line 37: Why was the QCD scale removed?

kmohrman: For ttgamma, we don't have an NLO xsec for this sample, so we're using the LO one from the MG gridpack. Since it's an LO xsec, we would want to calculate the uncertainty via the renorm/fact and ISR/FSR systematics (if I'm understanding correctly the explanation from @klannon). Usually (when we have an NLO xsec) for these uncertainties do this (e.g. for the up variation):
`xsec_NLO*(N_pass_up/N_gen_up)`
As this allows us to account for the shape differences, while leaving the xsec alone (since the uncertainty on that is taken care of with the rate uncertainty in the datacard). However, in the case where we have an LO xsec, we can do this:
`xsec_NLO*(N_pass_up/N_gen)`
As this will include both the shape and rate uncertainties. Since we're estimating the uncertainty on the LO xsec in this way, we would not want to also apply the NLO xsec uncertainty on top of that. 

I've attempted to also explain this in the processor [here](https://github.com/TopEFT/topcoffea/blob/27d1c0e25eef80a59f042e57757294d962c6a76f/analysis/topEFT/topeft.py#L126-L154) where I've modified the up/down sum of weights to be the nominal sum of weights (in the case of the ttgamma LO samples).

kmohrman: Right, I think so.

kmohrman:
Additional checks have shown that the way we are handling the LO xsec uncertainty for the LO ttgamma samples seems to be consistent with what we intended.

Running this through the full workflow, the limits change by ~1% or less compared to the case where we do not include conversions at all.

I think this branch is ready to be merged, unless @bryates or anyone has any comments or questions or objections.

bryates:
> Additional checks have shown that the way we are handling the LO xsec uncertainty for the LO ttgamma samples seems to be consistent with what we intended.
> 
> Running this through the full workflow, the limits change by ~1% or less compared to the case where we do not include conversions at all.
> 
> I think this branch is ready to be merged, unless @bryates or anyone has any comments or questions or objections.

Thanks for the careful checks, it's good to see the effects are small. I have no objections to merging now.

PR #275, opened 2022-05-26
Fixing triggers

sscruz:
Fixes the trigger selection, since there isn't a `DoubleEG` dataset in 2018, all the single and electron triggers are in `EGamma`


bryates (approved):

kmohrman:
Just for the record, this update successfully fixes the 2018 discrepancy we observed in the 2los Z CR (specifically for ee).

PR #276, opened 2022-05-27
Updated PDF/QCD rates

bryates:
This PR only updates the rate uncertainties where I've found confirmed better ones. The only exceptions are for `tHq`:
1. PDF $\to$ quad sum of 3.7% and $\alpha_s$ 3.5%
1. QCD $\to$ -14.9%/+6.5% (this was used in TOP-19-001, but was cited wrong)


kmohrman:
Looks good to me.

PR #278, opened 2022-06-02
Central signal jsons

kmohrman:
This PR adds JSONs for central ttH, ttW, ttZ, and 4t samples. 

The JSONs have been added and the sum of weights have been calculated, but note that these samples still need to be skimmed before we merge.


bryates (approved):
Looks good, thanks for adding these.

Andrew42:
I've added the JSON files for the skimmed versions of the central signal samples.

As an archival note, the lobster skim job also included the tZq central samples, for which we had already produced skims. After talking with @kmohrman it was decided to use the older tZq skims for the sake of consistency, however, either variant of the tZq skim should be identical in terms of event content as the lobster job simply ran over the same un-skimmed tZq samples as before.

I've also added a new `ND_central_sgnl_mc_skim_locations.txt` to indicate which skims are actually used in making the JSON files, as well as `mc_central_signal_samples_NDSkim.cfg` for easily running over these new skims as a single group.

The `python` command used to generate the JSON skim files was:
```
python make_skim_jsons.py --file ND_central_sgnl_mc_skim_locations.txt --json-dir ../../topcoffea/json/signal_samples/central_UL
```

PR #282, opened 2022-06-08
Fix for the HistEFT rebin method to preserve bin array shape

Andrew42:
Fixes a bug in `HistEFT` where if you call the `rebin()` method. Bins without an EFT parameterization were getting converted into a form as if they were meant to have an EFT parameterization. Basically 1D arrays were getting turned into 2D arrays. Below is a MRE that should better illustrate the point:
```python
from coffea.hist import StringBin, Cat, Bin
from topcoffea.modules.HistEFT import HistEFT

wcnames = ["ctG"]
h = HistEFT("Events",wcnames,Cat("sample","sample"),Bin("pt","pt",5,0,5))

coeffs = [
    [1,2,3],
]

h.fill(sample="EFT",pt=1.0,eft_coeff=coeffs)
h.fill(sample="EFT",pt=2.0,eft_coeff=coeffs)
h.fill(sample="EFT",pt=5.0,eft_coeff=coeffs)

h.fill(sample="SM",pt=1.0,eft_coeff=None)
h.fill(sample="SM",pt=2.0,eft_coeff=None)
h.fill(sample="SM",pt=5.0,eft_coeff=None)

new_edges = [0,3,5]
h_rebin = h.rebin("pt",Bin("pt","pt",new_edges))

print("-"*50)
print("- Before calling .rebin()")
print("-"*50)
for k,v in h._sumw.items():
    print(k); print(v);
print("")
print("-"*50)
print("- After calling .rebin()")
print("-"*50)
for k,v in h_rebin._sumw.items():
    print(k); print(v);
```
The fix was to make sure that shape of the dense axis is preserved during rebinning.

### DIFF
--- a/topcoffea/modules/HistEFT.py
+++ b/topcoffea/modules/HistEFT.py
@@ -643,12 +643,6 @@ class HistEFT(coffea.hist.Hist):
         anew[view_ax(inew)] += array[view_ax(iold)]
       return anew
 
-    # def dense_op(array):
-    #   anew = np.zeros(shape=(*out._dense_shape,out._ncoeffs), dtype=out._dtype)
-    #   for iold, inew in enumerate(binmap):
-    #     anew[view_ax(inew)] += array[view_ax(iold)]
-    #   return anew
-
     for key in self._sumw:
       out._sumw[key] = dense_op(self._sumw[key])
       if self._sumw2 is not None:
--- a/topcoffea/modules/HistEFT.py
+++ b/topcoffea/modules/HistEFT.py
@@ -633,10 +633,21 @@ class HistEFT(coffea.hist.Hist):
     binmap = [new_axis.index(i) for i in old_axis.identifiers(overflow='allnan')]
 
     def dense_op(array):
-      anew = np.zeros(shape=(*out._dense_shape,out._ncoeffs), dtype=out._dtype)
+      if array.shape != self._dense_shape:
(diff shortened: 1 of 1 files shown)


bryates (approved):
Thanks @Andrew42! Do you want to add your example code to the pytest scripts, or is it not worth repeatedly running a test on the rebin function?

Andrew42:
> Thanks @Andrew42! Do you want to add your example code to the pytest scripts, or is it not worth repeatedly running a test on the rebin function?

My MRE isn't really a test. It would need to be modified to actually check for what goes wrong. I _guess_ you would just need to add a check to make sure the shapes of the arrays are what you expect them to be, but I don't have time to play around with it right now and make sure it actually tests what we want it to test.

bryates:
> > Thanks @Andrew42! Do you want to add your example code to the pytest scripts, or is it not worth repeatedly running a test on the rebin function?
> 
> My MRE isn't really a test. It would need to be modified to actually check for what goes wrong. I _guess_ you would just need to add a check to make sure the shapes of the arrays are what you expect them to be, but I don't have time to play around with it right now and make sure it actually tests what we want it to test.

That's ok, it's not really _needed_.

kmohrman:
Thanks very much @Andrew42. I'm happy to see that the CI is passing on this branch, as I guess that implies this bug at least does not affect anything we're testing with the CI. 

I'd be happy to merge it now, unless anyone else wants to look at it in more detail first.

PR #286, opened 2022-06-10
decrease max backoff of wq worker in test

btovar:
This should help eliminate the wq test timing-out, as the worker won't
wait for a long time between retries contacting the master.


bryates (approved):

kmohrman:
Thanks very much for this update, Ben. It looks good to me.

PR #289, opened 2022-06-11
Renormfact envelope

kmohrman:
This PR adds a script to calculate the renorm/fact envelope from the 6 renorm and fact variations. It is called from the run script after the data driven estimator is run (taking as input a pkl file, and writing the output to the same pkl file, where in each bin the 6 renorm and fact variations have been replaced with just the two most extreme variations)

As discussed on MM, the effect on the limits of implementing this envelope is quite small (about 2% at the largest).


bryates (approved):
Looks good to me, thanks @kmohrman!

PR #290, opened 2022-06-12
MC validation scripts

kmohrman:
This PR adds a few scripts that were used during the validation studies of the private MC (which we performed in June of 2022 as part of the pre-approval studies and checks suggested during the pre-approval presentation). The main scripts added are the following:
* `mc_validation_plotter.py`: This is the plotting script we used for most of the private vs central RECO comparisons. The calculation of the systematic errors (on the private samples) is included. 
* `mc_validation_gen_processor.py`:  This is a simple processor that fills GEN-level distributions in order to compare e.g. private to central (either UL or pre-UL). 
* `mc_validation_gen_plotter.py`: This simple plotting script runs on the output of the GEN processor mentioned above.

I'm not sure if these scripts will actually be useful for anything in the future, but I wanted to get them committed just in case.


bryates (approved):
Looks good, thanks for adding these.

PR #302, opened 2022-07-03
Check to make sure selected WCs agree

kmohrman:
Add a check to the data card maker wrapper to raise and error if the list of selected WCs does not match the reference. Note this is not really the best way to handle this check (would be better to check against the one in the EFTFit repo, but that's harder) but at least should be better than nothing.


topcoffea/modules/utils.py
```diff
@@ -162,3 +162,27 @@ def get_hist_from_pkl(path_to_pkl,allow_empty=True):
...
+# Check if the contents of two dictionaries of lists agree
+# Assumes structure d = {k1: [i1,i2,...], ...}
+def dict_comp(in_dict1,in_dict2,strict=False):
+
+    def all_d1_in_d2(d1,d2):
+        agree = True
+        for k1,v1 in d1.items():
+            if k1 not in d2:
+                agree = False
+                continue
```
bryates, line 175: Probably won't affect performance, but you could put `break` here, or even `return False`.

topcoffea/modules/utils.py
```diff
@@ -162,3 +162,27 @@ def get_hist_from_pkl(path_to_pkl,allow_empty=True):
...
+            for i1 in v1:
+                if i1 not in d2[k1]:
+                    agree = False
```
bryates, line 178: Same here

bryates (approved):
Looks good. I had a small comment that doesn't need to be added.

topcoffea/modules/utils.py
kmohrman: Thanks, added a `break`.

topcoffea/modules/utils.py
kmohrman: Thanks, added a `break`.

PR #305, opened 2022-07-05
Increase text datacard precision to 6 decimal places

Andrew42:
Changes the decimal precision output by the datacard maker for the text datacard.

### DIFF
--- a/topcoffea/modules/datacard_tools.py
+++ b/topcoffea/modules/datacard_tools.py
@@ -14,7 +14,7 @@ from coffea.hist import StringBin, Cat, Bin
 from topcoffea.modules.paths import topcoffea_path
 import topcoffea.modules.eft_helper as efth
 
-PRECISION = 3   # Decimal point precision in the text datacard output
+PRECISION = 6   # Decimal point precision in the text datacard output
 
 def prune_axis(h,axis,to_keep):
     """ Convenience method to remove all categories except for a selected subset."""


bryates (approved):
Looks good to me.

kmohrman:
The CI seems to be failing

Andrew42:
I ran these particular CI tests by hand and what I get for the offending column (`ttH_quad_ctp`) is the following:
```
f_new = 0.037179
f_ref = 0.037
(f_new - f_ref) / f_ref = 0.0048378378378378
```
So it looks like the difference caused by the rounding in the ref file was enough to trigger the CI to fail this test. I should also point out that when I ran the CI test myself it actually complained about a different column (`ttH_quad_mixed_ctq8_cQq13`), which corresponds to:
```
f_new = 12.877915
f_ref = 12.878
(f_new - f_ref) / f_ref = -6.600403789434291e-06
```
Given the nature of the change in this PR, I don't think we wold consider these differences as bad, but I didn't write the CI tests for this particular check.

kmohrman:
Sounds reasonable to me. Can your remake the reference?

Andrew42:
I don't know how the reference was made originally. I could commit the output I got from running the CI tests by hand, but I _only_ ran the tests associated with the {code}test_futures.py{code} file, so I don't know if the other tests or CI steps would modify things that I need to also account for? Like I said, when I ran the tests by hand it actually threw an error for a _different_ column than what the CI complained about here.

kmohrman:
You just need to remake the reference text file in the analysis/topEFT/test dir. The `remake_ci_ref_datacard.sh` can be used for that.

PR #306, opened 2022-07-07
Minor updates

kmohrman:
This PR includes two minor, unrelated updates:
* Edit the logic in the datacard maker's selected WCs check so that the check is skipped in the case that we are providing the script with a set of WCs.
* Fix the names in the channel name map in the `get_datacard_yields.py` script to match the category names in the new datacard maker.


bryates (approved):
Looks good.

Andrew42:
looks good to me as well!

Andrew42 (approved):

PR #308, opened 2022-07-08
Add comparison check for blind Observed yields

kmohrman:
This PR adds a check to the `get_datacard_yields.py` so that we can compare the Observed yields to the sm predicted yields, and raise a warning if they differ by a large factor. Also added and `--unblind` option, now it this option is not specified the Observed yield will be reported as -999.


bryates (approved):
Looks good, thanks for adding this check!

PR #310, opened 2022-07-12
Fix datacard maker histogram bin errors

Andrew42:
This PR fixes a missed inconsistency with the old datacard maker, which was for the signal process histograms the bin errors were explicitly set to 0, while for the background processes the bin error was left unset.

### DIFF
--- a/topcoffea/modules/datacard_tools.py
+++ b/topcoffea/modules/datacard_tools.py
@@ -28,7 +28,10 @@ def to_hist(arr,name,zero_wgts=False):
         uproot. If 'zero_wgts' is true, then the resulting histogram will be created with bin errors
         set to 0 (instead of left unset)
     """
-    clipped = arr[1:-1]     # Strip off the under/overflow bins
+    # NOTE:
+    #   If we don't instantiate a new np.array here, then clipped will store a reference to the
+    #   sub-array arr and when we modify clipped, it will propagate back to arr as well!
+    clipped = np.array(arr[1:-1])     # Strip off the under/overflow bins
     clipped[-1] += arr[-1]  # Add the overflow bin to the right most bin content
     nbins = len(clipped)
     if zero_wgts:
--- a/topcoffea/modules/datacard_tools.py
+++ b/topcoffea/modules/datacard_tools.py
@@ -2,6 +2,7 @@ import pickle
 import gzip
 import topcoffea.modules.HistEFT
 import numpy as np
+import boost_histogram as bh
 import uproot
 import hist
 import os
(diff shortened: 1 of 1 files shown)


topcoffea/modules/datacard_tools.py
```diff
@@ -2,6 +2,7 @@
 import gzip
 import topcoffea.modules.HistEFT
 import numpy as np
+import boost_histogram as bh
```
bryates, line 5: Is `bh` actually used?

bryates (approved):
Thanks @Andrew42, this looks good. I just had one small question.

topcoffea/modules/datacard_tools.py
Andrew42: yeah, it's needed to specify the `storage` type of the `hist` histogram when we want to explicitly tell it to save bin weights. So this comes up [here](https://github.com/TopEFT/topcoffea/blob/0d3a827c2bea94c42663d38e8d3ad0508f6dc501/topcoffea/modules/datacard_tools.py#L35).

bryates: Ah thanks, I didn't scroll the window far enough to see it!

Andrew42:
@kmohrman noticed a subtle bug introduced into the `to_hist()` function. I was creating a new variable like so:
```python
arr = np.array([1,2,3,4])
clipped = arr[1:-1]
```
I had thought (based on my understanding of python slice syntax) that this would create a copy of the sub-array specified by the slice. This, however, is not the case (at least for numpy arrays)! As a result, later on when I merge the overflow bin into the right most bin of `clipped` like so:
```python
clipped[-1] += arr[-1]
```
This actually modifies the `arr` array. Namely, it increments the 2nd to last element of `arr` by w/e the last element of `arr` is.

Interestingly, when we set the bin contents of the `hist.Hist` object I had made a 2nd mistake (also fixed now) where I was still using the `arr` array to set the bin contents for the signal samples instead of the `clipped` array. This 2nd mistake just so happened to exactly cancel the 1st mistake, which is why I probably didn't notice this issue when testing my changes originally.

kmohrman:
Sorry for being slow to finish the review on this PR. I think our checks over the past couple weeks have confirmed that the auto-stats issues we are seeing with the unblinded workspace are probably not related to any issues with this PR. Running with asimov data and auto-stats on (a check we had done a couple weeks ago) we saw no issues. So it would probably be good to merge this PR now so we do not have to keep making workspaces on this branch (and can just use the master instead). I see Brent already approved this PR as well, so I'll go ahead and merge now.

PR #314, opened 2022-08-07
Add TWZ bkg JSONs

kmohrman:
This PR adds new JSONs (with sum of weight information included) for the TWZ background. This branch should not be merged until the samples are skimmed and skimmed JSONs are added and tested (to make sure the yields look as expected).


bryates (approved):

Andrew42:
The tWZ skims have finally finished and I've added the corresponding JSON files here. The json files were regenerated using:
```
```
I also added a commit that updates the `TTGamma` json files which removes an EOF newline character that got introduced at some point. This doesn't actually impact the topeft processor in any way, but does ensure that running `make_skim_jsons.py` doesn't continue to generate spurious `git diff` results after running.

bryates (approved):

kmohrman:
As mentioned on [MM](https://mattermost.web.cern.ch/topeft-runii/pl/w4t7xkpb6pga3y8sip1unkxjfh), the new tWZ samples look good, so I think we should be ready to merge. 

However, unrelated to the original PR, I've also updated the ttZ and tttt cross sections (in the cfg file, the unskimmed JSONs, and the skimmed JSONs). I'm wondering if @bryates and/or @Andrew42 could take a look to make sure that all looks ok.

bryates (approved):
Thanks for taking care of the xsecs @kmohrman, I think everything looks good. I can't remember now why we stored them in the json files instead of always just reading from the `xsec.cfg`, but maybe there was a good reason.

kmohrman:
Ok, thanks for checking @bryates. Yeah, I completely agree that it would be much better to just read the values from the xsec config in all cases. Something to implement for the next iteration, I guess.

Andrew42:
I don't see anything wrong with the changes either, so looks good to me

PR #315, opened 2022-08-09
Fix triggerSF lepton flavor selection logic

Andrew42:
It was noticed while reviewing the code that during triggerSF application, the criteria for identifying lepton flavor could have overlaps between 2l and 3l events. For example, the criteria `event.is_eee` being true would also mean that `event.is_ee` is true, which wasn't being accounted for in the triggerSF corrections. This meant that certain events had the triggerSF applied twice.

This PR simply adds the additional requirement `is2l` and `is3l` in their corresponding use cases. The overall effect of this change is pretty small and is largest in the 3l categories. I will follow-up with yield tables that show the changes before and after applying the fix.


Andrew42:
The yield tables showcasing the impact of the fix [before](https://www.crc.nd.edu/~awightma/github/TopEFT/PR/315/before_fix_v1.pdf) and [after](https://www.crc.nd.edu/~awightma/github/TopEFT/PR/315/after_fix_v1.pdf)

bryates (approved):
Thanks for fixing this.

Andrew42:
I guess it is worth pointing out that with this fix we no longer apply the triggerSF corrections to 4l events. However we don't have 4l triggerSFs to begin with. In principle I guess we could just use the 3l SFs, which is probably most similar to what was being done before this fix, but it then raises the question about which flavor combination to use and how to avoid double applying the SF.

@andrea21z do you have any thoughts on this?

kmohrman:
As described in Issue #316, it probably makes sense to go ahead and merge this PR now, even though the 4l SFs will still need to be discussed (and possibly a new PR will have to be made once we know how they should be handled, but it would still be good to get this PR merged now, so that at least the 3l SFs will be correct).

PR #321, opened 2022-09-09
added jsons for new ZZ sig samples and ran update sow on them

abasnet97:
In this PR, I have added JSONS. for the new ggToZZ* samples from AN-19-156. I added the cross-section values for the corresponding signal processes and then also updated the sum of weights of the JSONs.


bryates (approved):
Looks good!

Andrew42:
The skims just finished and I've created the json files for the newly skimmed samples. I added these new samples to `ND_bkgd_mc_skim_locations.txt` and the corresponding command used to regenerate all associated json files was:
```
```

kmohrman:
I think we should go ahead and merge this branch, as all of the updates have been checked (and included in the results shown at the Sep 12, 2022 unblinding [update](https://indico.cern.ch/event/1199675/contributions/5044415/attachments/2506112/4307218/top22006-update_byates_9-12-22.pdf) of TOP-22-006 at the TMP meeting). Beyond just the ZZ samples, this branch ended up including a couple other updates too:
* Setting sumw2 to 0 for all but the fakes (so only fakes are included for auto stats)
* Including the new samples in the list of samples for prompt subtraction (including the central signal samples as well, so that when we run with these for comparison, the fakes are right) 

Unless there are any objections, I'll go ahead and merge.

PR #323, opened 2022-09-13
Fix statistical uncertainty in datacards and crop negative bins to zero

sscruz:
- Fixes the bug that was setting all statistical uncertainties to `sqrt(bin content)` 
- Crops negative bins to zero (affecting only fakes estimation)

### DIFF
--- a/topcoffea/modules/datacard_tools.py
+++ b/topcoffea/modules/datacard_tools.py
@@ -41,11 +41,10 @@ def to_hist(arr,name,zero_wgts=False):
 
 
     nbins = len(clipped[0])
+    h = hist.Hist(hist.axis.Regular(nbins,0,nbins,name=name),storage=bh.storage.Weight())
     if zero_wgts:
-        h = hist.Hist(hist.axis.Regular(nbins,0,nbins,name=name),storage=bh.storage.Weight())
         h[...] = np.stack([clipped[0],np.zeros_like(clipped[0])],axis=-1) # Set the bin errors all to 0
     else:
-        h = hist.Hist(hist.axis.Regular(nbins,0,nbins,name=name),storage=bh.storage.Weight())
         h[...] = np.stack([clipped[0], clipped[1]],axis=-1)
     return h
 
@@ -755,7 +754,7 @@ class DatacardMaker():
             print(f"WC Selection Time: {dt:.2f} s")
         return selected_wcs
 
-    def analyze(self,km_dist,ch,selected_wcs, cropNegativeBins):
+    def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins):
         """ Handles the EFT decomposition and the actual writing of the ROOT and text datacard files."""
         if not km_dist in self.hists:
             print(f"[ERROR] Unknown kinematic distribution: {km_dist}")
@@ -782,7 +781,7 @@ class DatacardMaker():
 
         h = self.hists[km_dist]
         ch_hist = h.integrate("channel",[ch])
-        data_obs = [np.zeros(ch_hist._dense_shape[0] - 1), np.zeros(ch_hist._dense_shape[0] - 1)] # '_dense_shape' includes the nanflow bin
+        data_obs = np.zeros((2,ch_hist._dense_shape[0] - 1)) # '_dense_shape' includes the nanflow bin
 
         print(f"Generating root file: {outf_root_name}")
         tic = time.time()
@@ -806,7 +805,7 @@ class DatacardMaker():
                         elif sum(data_obs[0]) != 0:
                             raise RuntimeError("filling obs data more than once!")
                         for sp_key,arr in data_sm.items():
-                            data_obs[0] += arr[0]
+                            data_obs += arr
                 for base,v in decomposed_templates.items():
                     proc_name = f"{p}_{base}"
                     col_width = max(len(proc_name),col_width)
@@ -816,7 +815,7 @@ class DatacardMaker():
                     }
                     # There should be only 1 sparse axis at this point, the systematics axis
                     for sp_key,arr in v.items():
-                        if cropNegativeBins:
+                        if crop_negative_bins:
                             negative_bin_mask = np.where( arr[0] < 0) # see where bins are negative
                             arr[0][negative_bin_mask] = np.zeros_like( arr[0][negative_bin_mask] )  # set those to zero
                             if arr[1] is not None:
@@ -830,7 +829,7 @@ class DatacardMaker():
                                 print(f"\t{proc_name:<12}: {sum_arr:.4f} {arr[0]}")
                             if not self.use_real_data:
                                 # Create asimov dataset
-                                data_obs[0] += arr[0]
+                                data_obs += arr
                         if syst == "nominal":
                             hist_name = f"{proc_name}"
                             text_card_info[proc_name]["rate"] = sum_arr
--- a/topcoffea/modules/datacard_tools.py
+++ b/topcoffea/modules/datacard_tools.py
@@ -22,7 +22,7 @@ def prune_axis(h,axis,to_keep):
     to_remove = [x.name for x in h.identifiers(axis) if x.name not in to_keep]
     return h.remove(to_remove,axis)
 
-def to_hist(arr,name,zero_wgts=False):
+def to_hist(arr,name,zero_wgts=False, cropNegativeBins=False):
     """
         Converts a numpy array into a hist.Hist object suitable for being written to a root file by
         uproot. If 'zero_wgts' is true, then the resulting histogram will be created with bin errors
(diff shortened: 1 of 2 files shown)


bryates (approved):
Looks good, thanks for catching and fixing this!

topcoffea/modules/datacard_tools.py
```diff
@@ -1016,13 +1029,15 @@ def decompose(self,h,wcs):
...
+                for i in range(2):
+                    r[quad_name][tup].append( 0.5*(tmp_lin_2[tup][i] - 2*tmp_lin_1[tup][i] + sm[tup][i]) ) 
+                    
             for n2,wc2 in enumerate(wcs):
                 if n1 >= n2: continue
                 mixed_name = f"quad_mixed_{wc1}_{wc2}"
                 h.set_wilson_coefficients(**{wc1:1.0,wc2:1.0})
-                r[mixed_name] = h.values(overflow='all')
+                r[mixed_name] = h.values(overflow='all',sumw2=True)
                 terms += 1
```
Andrew42, line 1041: This is a general comment for the changes in this entire function. I _think_ it's safe to drop the `i` indexing for `numpy` arrays and it will automatically handle the multiplication and addition by sub-array. It also has the benefit of keeping everything in a  `numpy`/`awkward` array, rather than introducing a (admittedly small) vanilla python `list`.

topcoffea/modules/datacard_tools.py
```diff
@@ -31,15 +31,22 @@ def to_hist(arr,name,zero_wgts=False):
...
+    nbins = len(clipped[0])
     if zero_wgts:
         h = hist.Hist(hist.axis.Regular(nbins,0,nbins,name=name),storage=bh.storage.Weight())
-        h[...] = np.stack([clipped,np.zeros_like(clipped)],axis=-1) # Set the bin errors all to 0
+        h[...] = np.stack([clipped[0],np.zeros_like(clipped[0])],axis=-1) # Set the bin errors all to 0
     else:
-        h = hist.Hist(hist.axis.Regular(nbins,0,nbins,name=name))
-        h[...] = clipped
+        h = hist.Hist(hist.axis.Regular(nbins,0,nbins,name=name),storage=bh.storage.Weight())
+        h[...] = np.stack([clipped[0], clipped[1]],axis=-1)
```
Andrew42, line 49: Since the instantiation of `h` is now identical in both branches of the `if` statement resolution, we should probably move it outside of the `if` block entirely.

topcoffea/modules/datacard_tools.py
```diff
@@ -31,15 +31,22 @@ def to_hist(arr,name,zero_wgts=False):
...
-    clipped = np.array(arr[1:-1])     # Strip off the under/overflow bins
-    clipped[-1] += arr[-1]  # Add the overflow bin to the right most bin content
-    nbins = len(clipped)
+    clipped=[]
+    for i in range(2): # first entry is sum(weight), second entry is sum(weight^2)
+        if arr[i] is not None:
+            clipped.append( np.array(arr[i][1:-1]))     # Strip off the under/overflow bins
+            clipped[i][-1] += arr[i][-1]  # Add the overflow bin to the right most bin content
+        else: 
+            clipped[i]=None
```
Andrew42, line 40: In the same vein as some of my later comments, I feel like this block of code can be made to be more `numpy`-like, but the array manipulations here are sufficiently complex that I would need to sit down and make sure to get the syntax right. As it is (mostly) just an aesthetic/styling change, I'm fine with skipping over it for now and revising it later.

Andrew42 (commented):
Here are some of the comments I had on the PR.

Pretty much all of them relate to code style and so shouldn't have any impact on the performance or implementation of the changes. I would like to make an effort to keep a consistent code style where possible, even if we haven't been very good about that in other parts of the code base.

Other than that, we might want to revisit this at some point to make the code only carry around the `sumw2` array for those samples we actually intend to have MC stat uncertainties included, but that would mainly be a performance optimization thing and shouldn't be something that stops us from moving forward on the analysis and so I'm fine with ignoring it for now.

topcoffea/modules/datacard_tools.py
```diff
@@ -796,10 +803,10 @@ def analyze(self,km_dist,ch,selected_wcs):
                     if self.use_real_data:
                         if len(data_sm) != 1:
                             raise RuntimeError("obs data has unexpected number of sparse bins")
-                        elif sum(data_obs) != 0:
+                        elif sum(data_obs[0]) != 0:
                             raise RuntimeError("filling obs data more than once!")
                         for sp_key,arr in data_sm.items():
-                            data_obs += arr
+                            data_obs[0] += arr[0]
```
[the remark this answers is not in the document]
Andrew42: for some reason my "above change" didn't get saved and posted. What I was referring to was the following line that declares the `data_obs` array:
```python
data_obs = [np.zeros(ch_hist._dense_shape[0] - 1), np.zeros(ch_hist._dense_shape[0] - 1)]
```
This can be changed to
```python
data_obs = np.zeros((2,ch_hist._dense_shape[0] - 1))
```
Which keeps `data_obs` as an `np.array` and allows for the more compact array addition syntax I mentioned here and elsewhere.

kmohrman:
@Andrew42 and @sscruz, I am wondering about the plan regarding these style comments. Do either (or both) of you want to (and have time to) implement these updates as part of this PR? Or is this something that will be pushed back to an issue to be addressed later?

Andrew42:
@kmohrman the style comments are pretty simple/minor, so I could do them myself, but as this isn't my PR I don't want to commit changes w/o @sscruz's acknowledgement or feedback

Andrew42:
@sscruz has implemented the requested changes and I'm fine with having this PR merged

Andrew42:
I ran a comparison of the datacard maker outputs from `master` (f3) and the changes from commit `0ceefc9` (f2) and compared against the outputs used to generate the most recent workspace, which should correspond to this PR prior to commit `2e8de7f` (f1):
```python
fpath_1 = "/scratch365/kmohrman/fullR2_files/anatest21v01/ptz-lj0pt_withSys"
fpath_2 = "/scratch365/awightma/datacard_tests/fixNaNuncertainties"
fpath_3 = "/scratch365/awightma/datacard_tests/master_61eb321d"

km = "lj0pt"
ch = "2lss_4t_m_4j"
fname = f"ttx_multileptons-{ch}_{km}.root"

f1 = uproot.open(os.path.join(fpath_1,fname))
f2 = uproot.open(os.path.join(fpath_2,fname))
f3 = uproot.open(os.path.join(fpath_3,fname))

h1 = f1["fakes_sm"]
h2 = f2["fakes_sm"]
h3 = f3["fakes_sm"]

h1.values() # array([1.23129159, 1.50385304, 0.16989791, 0.        ])
h2.values() # array([1.23129159, 1.50385304, 0.16989791, 0.        ])
h3.values() # array([1.23129159, 1.50385304, 0.16989791, -0.03409792])

h1.variances() # array([0.21787236, 0.25944065, 0.04309692, 0.        ])
h2.variances() # array([0.21787236, 0.25944065, 0.04309692, 0.        ])
h3.variances() # array([1.23129159, 1.50385304, 0.16989791, 0.        ])
```
As expected, the output from `master` generates bins with negative events, while using the version from this PR the bin is correctly zero'd out. Furthermore, the style changes did not affect the outputs (as expected).

I can run this comparison over more channels if we think it's necessary, but from this I'd say it would be safe to merge

sscruz:
hi, I ran a similar validation with the script below, getting also no changes in all bins of our datacards, besides the `fakes_sm`
```
import ROOT as r
import os

import re
pattern=re.compile('(?P<proc>.*)_quad_mixed_(?P<wc1>.*)_(?P<wc2>.*)')
for l in os.listdir('.'):
    if not l.endswith('.root'): continue
    if 'njet' in l: continue
    if '3l' in l and 'lj0pt' in l: continue
    tf=r.TFile.Open(l)
    alternative_tf=r.TFile.Open('/work/kmohrman/CMSSW_Releases/CMSSW_10_2_13/src/EFTFit/Fitter/test/datacards/fullR2/anatest19/at19v01/ptz-lj0pt_withSys/%s'%l)
    if not alternative_tf:
        continue
    for key in  tf.GetListOfKeys():
        hist = key.ReadObj()
        hist_name=hist.GetName()
        if not alternative_tf.Get(hist_name):
            # try something                                                                                                                                                                                        
            match = pattern.match(hist_name)
            if  match:
                hist_name = '%s_quad_mixed_%s_%s'%(match.group('proc'), match.group('wc2'), match.group('wc1'))

        if not alternative_tf.Get(hist_name):
            print("Histo %s not found in file %s"%(hist_name, l))
            print('the integral is', hist_name)
            continue
        alt_hist=alternative_tf.Get(hist_name)

        for i in range(1,hist.GetNbinsX()+1):
            if abs(alt_hist.GetBinContent(i)-hist.GetBinContent(i)) < 1e-10: continue
            print(hist_name, alt_hist.GetBinContent(i)-hist.GetBinContent(i))
```

kmohrman:
Thank you @sscruz and @Andrew42. It looks like all of the comments on this PR have been addressed, and that everything has been double checked by both of you. Also @bryates has approved the PR. So I will go ahead and merge.

PR #324, opened 2022-09-19
Handle non eft cases in copy_sm()

kmohrman:
This PR adds a check so that non-EFT samples can be handled in the `copy_sm()` method. This does not come up in most of our use cases, but I did run into it a few weeks ago while running the non prompt estimation on a pkl file that had central signal samples. The error was this:
```
  File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/check_PRs/kmohrman_histeft/topcoffea/analysis/topEFT/do_np.py", line 12, in <module>
    ddp = DataDrivenProducer(out_pkl_file,out_pkl_file_np)
  File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/check_PRs/kmohrman_histeft/topcoffea/topcoffea/modules/dataDrivenEstimation.py", line 22, in __init__
    self.DDFakes()
  File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/check_PRs/kmohrman_histeft/topcoffea/topcoffea/modules/dataDrivenEstimation.py", line 129, in DDFakes
    hPromptSub = hPromptSub.copy_sm()
  File "/afs/crc.nd.edu/user/k/kmohrman/coffea_dir/check_PRs/kmohrman_histeft/topcoffea/topcoffea/modules/HistEFT.py", line 195, in copy_sm
    out._sumw[sparse_key]=self._sumw[sparse_key][:,0]
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```
The updates in this PR works around the error. Probably I should have committed this somewhere a few weeks ago when I first ran into it, but I never did, and sort of forgot about the changes, but came across them recently while cleaning up my repository, so I figured it might be useful to commit them.

However, I'm certainly not an expert on HistEFT and have not modified it as much as e.g. @Andrew42, so @Andrew42 (and @bryates, and anyone else who might want to take a look), it would be great if you could check this PR for anything that looks suspicious or wrong. Thanks

### DIFF
--- a/topcoffea/modules/HistEFT.py
+++ b/topcoffea/modules/HistEFT.py
@@ -194,9 +194,9 @@ class HistEFT(coffea.hist.Hist):
     for sparse_key in self._sumw.keys():
       is_eft_bin = (self._sumw[sparse_key].shape != self._dense_shape)
       if is_eft_bin:
-        out._sumw[sparse_key]=self._sumw[sparse_key][:,0]
+        out._sumw[sparse_key] = np.copy(self._sumw[sparse_key][:,0])
       else:
-        out._sumw[sparse_key]=self._sumw[sparse_key]
+        out._sumw[sparse_key] = np.copy(self._sumw[sparse_key])
       out._sumw2[sparse_key]=np.zeros_like(out._sumw[sparse_key])
     return out
 
--- a/topcoffea/modules/HistEFT.py
+++ b/topcoffea/modules/HistEFT.py
@@ -192,7 +192,11 @@ class HistEFT(coffea.hist.Hist):
     out._sumw = {}
     out._sumw2 = {}
     for sparse_key in self._sumw.keys():
-      out._sumw[sparse_key]=self._sumw[sparse_key][:,0]
+      is_eft_bin = (self._sumw[sparse_key].shape != self._dense_shape)
+      if is_eft_bin:
+        out._sumw[sparse_key]=self._sumw[sparse_key][:,0]
+      else:
+        out._sumw[sparse_key]=self._sumw[sparse_key]
       out._sumw2[sparse_key]=np.zeros_like(out._sumw[sparse_key])
     return out
 


topcoffea/modules/HistEFT.py
```diff
@@ -192,7 +192,11 @@ def copy_sm(self):
     out._sumw = {}
     out._sumw2 = {}
     for sparse_key in self._sumw.keys():
-      out._sumw[sparse_key]=self._sumw[sparse_key][:,0]
+      is_eft_bin = (self._sumw[sparse_key].shape != self._dense_shape)
+      if is_eft_bin:
+        out._sumw[sparse_key]=self._sumw[sparse_key][:,0]
+      else:
+        out._sumw[sparse_key]=self._sumw[sparse_key]
```
Andrew42, line 199: You might want to check whether or not L199 creates a fresh array or if it just saves a reference to an underlying array and whether that's intended or not. For example, if you take the return of this method and modified it in some way, does that propagate back to the original:
```python
bin_idx = 1 # or w/e might be a valid value
sparse_key = ("some","valid","key")

orig_hist._sumw[sparse_key][bin_idx] = 125
sm_copy = orig_hist.copy_sm()
sm_copy._sumw[sparse_key][bin_idx] = 42
assert(sm_copy._sumw[sparse_key][bin_idx] != orig_hist._sumw[sparse_key][bin_idx])
```

Andrew42: @kmohrman sorry for taking so long to get around to testing this, but I think this isn't doing what you want it to do.

Below is a MRE that tries to showcase the problem:
```python
import numpy as np
from coffea import hist
from topcoffea.modules.HistEFT import HistEFT

nevts = 3
sparse_key = (hist.StringBin("ch01"),)
dense_idx  = 1 # Note: bin 0 is for the underflow
dense_vals = np.full(nevts,0.5) # [0.5 0.5 0.5]

orig_hist = HistEFT("events",[],hist.Cat("ch","ch"),hist.Bin("x","x-axis",1,0,1))
orig_hist.fill(ch=sparse_key[0],x=dense_vals)

# Check that we filled the histogram correctly
assert(orig_hist._sumw[sparse_key][dense_idx] == nevts) # --> OK

sm_copy = orig_hist.copy_sm()

# Check that the copied histogram has the correct bin value
assert(sm_copy._sumw[sparse_key][dense_idx] == nevts)   # --> OK

# Fill the copied histogram with more data
sm_copy.fill(ch=sparse_key[0],x=dense_vals)

assert(sm_copy._sumw[sparse_key][dense_idx] == 2*nevts) # --> OK

# Compare the bin content of the original and copied histogram
assert(sm_copy._sumw[sparse_key][dense_idx] != orig_hist._sumw[sparse_key][dense_idx])  # --> Fails
```

Basically, the underlying issue is that both `orig_hist` and `sm_copy` are storing a reference to the same underlying array. So when you modify the content of `sm_copy` it also modifies the content of `orig_hist`. I think this also impacts the eft case as well as showcased below:
```python
import numpy as np

orig_arr = np.array([[1,2,3],[4,5,6],[7,8,9]])
copy_arr = orig_arr[:,0]    # [1 4 7]
copy_arr[1] = 42

assert(copy_arr[1] != orig_arr[1,0])    # --> FAILS
```


I went and checked that changing:
```python
if is_eft_bin:
    out._sumw[sparse_key]=self._sumw[sparse_key][:,0]
else:
    out._sumw[sparse_key]=self._sumw[sparse_key]
```
to
```python
if is_eft_bin:
    out._sumw[sparse_key]=np.copy(self._sumw[sparse_key][:,0])
else:
    out._sumw[sparse_key]=np.copy(self._sumw[sparse_key])
```
seems to resolves the issue, but keep in mind that `np.copy` only does a "shallow" copy, so in the case where the "elements" of the `np.array` are themselves objects, then the copied array will still store a reference to those same objects. I _think_ this isn't an issue for us as all of the arrays in our histograms should just be simple numbers, but would be good to get others opinions on this if possible.

kmohrman:
Thank you for the careful checks @Andrew42. If I am understanding correctly, this issue is not specifically related to the new code in this PR, but is general to the copy_sm() method. 

Also, as far as I know, this function is only used in one place (in `dataDrivenEstimation.py`) [here](https://github.com/TopEFT/topcoffea/blob/histeft-copy-sm-handle-non-eft/topcoffea/modules/dataDrivenEstimation.py#L129). Since we're doing `hPromptSub = hPromptSub.copy_sm()` I think the issue you identified is no causing a problem here (luckily). 

However, if you want to push the fix for this issue, that is good with me (so that we can make sure this does not become an issue in the future).

topcoffea/modules/HistEFT.py
Andrew42: fixed applied in commit `aefee7a`

kmohrman:
Thanks for this fix @Andrew42. I've pulled the changes and run a pkl file though the data driven estimator with and without this update, and ended up with the same numbers in the output file. I think this is expected since the bug you fixed did not happen to affect this use case. So since things seem to look consistent, I will merge the PR.

PR #325, opened 2022-09-21
Do not merge: Additional histos from unblinding cross checks

kmohrman:
This purpose of this PR is to document the extra histograms and variables we included in the processor while performing the TOP-22-006 unblinding cross checks during July and August 2022. The extra things we added during these studies that we do not want to merge into the master are being committed to this `additional-histos` branch. Since there is not really a way to document a brach on GH, we are documenting this via this PR. 

There were many histograms that are added, but likely we will not need these in the master, so for now we are just leaving them on this `additional-histos` branch. For this reason, the `additional-histos` should not be deleted (right now anyway), as it is serving as a reference for these extra histograms, should we need them again during TOP-22-006.

### DIFF
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -4,6 +4,7 @@ import cloudpickle
 import json
 import pprint
 import copy
+import sys
 import coffea
 import numpy as np
 import awkward as ak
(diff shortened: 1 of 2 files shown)


kmohrman:
Closing the PR, as we do not intend to merge it for now.

PR #326, opened 2022-10-06
Fixing trigger s fs

andrea21z:
We have to update the  trigger SFs adding a cut in MET>120GeV.
--> 3lepton trigger SFs =~ 1.


kmohrman:
I think we can go ahead and merge this PR (as we discussed at the meeting this morning, the effect on the results is very small ~1%). Unless there were any additional checks that we wanted to do, I'll go ahead and merge now.

bryates:
> I think we can go ahead and merge this PR (as we discussed at the meeting this morning, the effect on the results is very small ~1%). Unless there were any additional checks that we wanted to do, I'll go ahead and merge now.

I'm fine with merging now.

PR #327, opened 2022-12-08
Make renormalization & factorization independent per-process systematics

Andrew42:
This PR adds a hardcoded check for the systematics `renorm` and `fact` and when it finds them it modifies the systematic name from `renorm` to `renorm_PROCESS` and similarly for the corresponding template histogram name `PROCESS_renormUp` to `PROCESS_renorm_PROCESSUp`.

For the processes that are split by EFT contribution, all decomposed terms will share a single nuisance parameter. So for example `ttll` may be split into `ttll_sm`,`ttll_lin_cpt`, `ttll_quad_cpt`, etc. There will be only one nuisance correlated over each of them (e.g. `renorm_ttll` or `fact_ttll`).

This PR also adds a new command-line option `--drop-syst` which can be used to specify one (or more) systematics that should be dropped from the pkl file before proceeding. Systematics that are dropped this way should match exactly what name of the systematic, however, the "Up"/"Down" postfix isn't needed as the code will remove both, unless you explicitly specify "SYSTNAMEUp", which will only drop the "Up" version of SYSTNAME. You can specify multiple systematics to drop at once by separating each name with by a space after the `--drop-syst` option.

### DIFF
--- a/analysis/topEFT/fullR2_run.sh
+++ b/analysis/topEFT/fullR2_run.sh
@@ -5,11 +5,11 @@ OUT_NAME="example_name"
 
 # Build the run command for filling SR histos
 CFGS="../../topcoffea/cfg/mc_signal_samples_NDSkim.cfg,../../topcoffea/cfg/mc_background_samples_NDSkim.cfg,../../topcoffea/cfg/data_samples_NDSkim.cfg"
-OPTIONS="--hist-list ana --skip-cr --do-systs -s 50000 --do-np --do-renormfact-envelope -o $OUT_NAME" # For analysis
+OPTIONS="--hist-list ana --skip-cr --do-systs -s 50000 --do-np -o $OUT_NAME" # For analysis
 
 # Build the run command for filling CR histos
 #CFGS="../../topcoffea/cfg/mc_signal_samples_NDSkim.cfg,../../topcoffea/cfg/mc_background_samples_NDSkim.cfg,../../topcoffea/cfg/mc_background_samples_cr_NDSkim.cfg,../../topcoffea/cfg/data_samples_NDSkim.cfg"
-#OPTIONS="--hist-list cr --skip-sr --do-systs --do-np --do-renormfact-envelope --wc-list ctG -o $OUT_NAME" # For CR plots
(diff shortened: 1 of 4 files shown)


bryates (approved):
Thanks @Andrew42, looks good to me!

kmohrman:
As discussed on MM and in the meetings, these changes (treating renorm and fact as independent systematics and and de correlating across processes) have been tested and the effects on the 2 sigma asimov profiled limits are very small (~1%). And we have agreed with the ARC and conveners to move to this new method of handing the renorm/fact systematics, so I will go ahead and merge the PR.

kmohrman:
Since we're treating renorm and fact independently, we no longer need to include the variation where they are fluctuated together.

Consequently, I have now updated the processor to no longer include the `renormfact` variations (i.e. the variations where renorm and fact are varied together). I have also updated the run script wrapper (`fullR2_run.sh`) to no longer pass the  `--do-renormfact-envelope` option to the run script.

@bryates I'm wondering if you could check that these changes look ok before we merge?

bryates:
> Since we're treating renorm and fact independently, we no longer need to include the variation where they are fluctuated together.
> 
> Consequently, I have now updated the processor to no longer include the `renormfact` variations (i.e. the variations where renorm and fact are varied together). I have also updated the run script wrapper (`fullR2_run.sh`) to no longer pass the `--do-renormfact-envelope` option to the run script.
> 
> @bryates I'm wondering if you could check that these changes look ok before we merge?

Yes, your changes all look good to me.

kmohrman:
> > Since we're treating renorm and fact independently, we no longer need to include the variation where they are fluctuated together.
> > Consequently, I have now updated the processor to no longer include the `renormfact` variations (i.e. the variations where renorm and fact are varied together). I have also updated the run script wrapper (`fullR2_run.sh`) to no longer pass the `--do-renormfact-envelope` option to the run script.
> > @bryates I'm wondering if you could check that these changes look ok before we merge?
> 
> Yes, your changes all look good to me.

Ok, thanks for checking. Once the CI finishes, I'll go ahead and merge.

PR #329, opened 2022-12-18
Update CR plotting code to un correlate renorm and fact across processes

kmohrman:
This PR updates the CR plotting code to decorrelate the renorm and fact systematics across processes. Some notes:
* This was prompted by the updated handling of the renorm/fact systematics suggested by the ARC
* Since we updated the way this is handled by the fit in the SR, we also should update in the CR
* Some caveats to this implementation in the CR:
    - With these updates, the code expects to get a pkl file that has separate `renorm` and `fact` systematics (and ideally no `renormfact` systematic, though it just skips this systematic if its present)
    - I've symmetrized the renorm and fact errors (so I guess this is different from a fully correct implementation..) 
    - I've generally decor related across processes, though left some processes as correlated, as we do in the SR (e.g. the diboson processes are grouped in the SR, so the renorm/fact systematics are correlated across WZ and ZZ so I've followed a similar approach here. Generally grouping across the processes that are plotted together in the histograms, but I've left the signal processes as un correlated (even though all signal is plotted together in one category) in an attempt to more closely match what the fit does)

### DIFF
--- a/analysis/topEFT/make_cr_and_sr_plots.py
+++ b/analysis/topEFT/make_cr_and_sr_plots.py
@@ -321,8 +321,6 @@ def get_rate_syst_arrs(base_histo,proc_group_map):
 # Wrapper for getting plus and minus shape arrs
 def get_shape_syst_arrs(base_histo):
 
-    print("\n\nHere in get_shape_syst_arrs")
-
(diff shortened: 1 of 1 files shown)


bryates (approved):
Looks good, thanks for implementing this!

PR #330, opened 2022-12-19
Rework Decorrelating Shape Systematics and Decorrelate ISR Systematic

Andrew42:
This PR expands the way shape systematics are decorrelated as initial introduced in PR #327. The way decorrelations are specified is now done via a dictionary defined in the `__init__()` method, which takes the following form:
```python
self.syst_shape_decorrelate = {
    "SYST_1": [
        {
            "matches": ["PROC_1","PROC_2"],
            "group": "GROUP_1"
        },
        {
            "matches": ["PROC_3","PROC_4"],
            "group": "GROUP_2"
        },
        ...
    ],
    "SYST_2": [
        ...
    ]
}
```
Where `matches` defines a list of strings that will match certain processes, which are to stay correlated for the corresponding systematic. The `group` entry is a string that is appended to the end of the systematic name and introduced as a separate systematic in the datacard.

If a systematic has a process that isn't matched to any of the groupings, then that process (as well as any other processes that did not get matched) will remain correlated and use the original systematic name, but if no such process(-es) remain after the groupings, the original systematic will no longer show up in the datacard outputs.

The code includes an explicit check to make sure that for a given systematic each process is matched _at most_ once, i.e. a process should not appear in more than one correlated grouping, since this would mean that the systematic for that process would end up being double counted.

In the special case where the `group` string is `""` (the empty string), then the logic changes to instead split _all_ processes matched via the corresponding `matches` list into fully decorrelated systematics. This is accomplished by creating a new systematic with a name that has the process named appended to the end. For example, `renorm` could be split up into `renorm_ttll`, `renorm_tllq`, etc.

It should be noted that any new systematics specified in this way does not actually introduce additional template histograms into the workspace. Instead, it takes all the template histograms for the specified systematic and renames them such that only the specified groupings (which may only contain a single process) stay correlated, rather than having all processes correlated for that systematic. This does however introduce new nuisance parameters into the resulting workspace.

In addition to reworking/expanding how decorrelating shape systematics is done, this PR also decorrelates the `ISR` systematic, such that only processes with the same production mode (e.g. QCD-induced vs. EWK-induced) are correlated. The new groupings are as follows:
```
group: gg
processes: ttH, ttll, tttt, convs

group: gq
processes: tHq

group: qq
processes: ttlnu, tllq, Diboson, Triboson
```
The MC process `tWZ` is the only remaining process with an `ISR` systematic not included in the above groupings and so keeps the original systematic name. However, since `tWZ` is the only un-grouped process it is effectively decorrelated from all other processes for `ISR`.

### DIFF
--- a/topcoffea/modules/datacard_tools.py
+++ b/topcoffea/modules/datacard_tools.py
@@ -13,6 +13,7 @@ import time
 from coffea.hist import StringBin, Cat, Bin
 
 from topcoffea.modules.paths import topcoffea_path
+from topcoffea.modules.utils import regex_match
 import topcoffea.modules.eft_helper as efth
 
 PRECISION = 6   # Decimal point precision in the text datacard output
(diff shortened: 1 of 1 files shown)


bryates (approved):
Thanks @Andrew42, this PR looks good to me.

kmohrman:
Just for future reference: As discussed on MM, this PR has been tested and the effect (of the updated ISR correlations) on the width of the 2 sigma profiled asimov limits is small (about half a percent at the most).

PR #333, opened 2023-02-28
do not check for git version when trying exclude patterns

btovar:
Now we simply try with and without patterns. There are some issues with setuptools.version where pkg_resources is not always available.


bryates (approved):
Looks good to me. @kmohrman I'm fine with merging this if you are.

kmohrman:
Thanks very much @btovar. I've tested this out and it solved the issue I was having with the env being rebuild unexpectedly. So I'll go ahead and merge it now.

Issue #343, opened 2023-02-28
Review and update READMEs


kmohrman:
Review all topcoffea README files, and make necessary updates. A file that is known to be especially outdated is the `topcoffea/analysis/topEFT` readme).

kmohrman:
Closed by PR #354. (Though of course the READMEs will need to be updated in the future as the repo changes, at least they are more or less mostly up to date at this point now.)

Issue #336, opened 2023-02-28
Remove unused code


kmohrman:
Review all topcoffea code (especially in the `topcoffea/analysis/topEFT` directory and the `topcoffea/topcoffea/modules` directory) for unused and outdated code.

kmohrman:
Addressed in PR #350.

kmohrman:
Closed by PR #354

PR #350, opened 2023-03-23
Remove unused code

kmohrman:
This PR removes scripts (e.g. plotting scripts, processors, etc) and files (e.g. SF files) that are outdated and unused. Some of the READMEs (mainly the one in analysis/topEFT) have also been updated to remove the description of the removed scripts (the the explanations for the remaining scripts have been somewhat reorganized). 

Addresses Issues #336 and #343.


bryates (approved):
Thanks @kmohrman for the tedious cleanup. It looks like all the removed files are indeed out of date. Have you run the full processor to make sure nothing current was accidentally removed?

kmohrman:
The CI runs, so it should be ok, but I can go ahead and run a full test run of the processor to make sure it's ok. By the way, I opened this PR against the `technical_improvements` branch not against master, as I think it'd probably be good to avoid messing with the master branch too much till TOP-22-006 is out the door. So would an alternative approach be to put off those final full sized tests till we have a few more of the technical modifications in place? 

I think it'd be good if most/all of the updates we make to address the technical improvements issues in the GitHub project are made to the `technical_improvements` branch for now. Then once TOP-22-006 is finalized we can merge in this branch (or if we finish with all of the technical improvements long before TOP-22-006 is finalized so we want to merge it into master, we will probably want to do a _lot_ of very thorough validation of the entire workflow to make sure everything looks ok beforehand).

kmohrman:
@bryates the full sized run of the processor just finished without any issues. And I also made cards from it (which also ran fine) and as a quick sanity check just verified that the observed number of events looked ok (and it indeed matched the right number).

kmohrman:
Ok, so since I think the run of the processor was your only comment (and that seemed to check out), and since you've approved the PR, and since we're not even merging into master right now, just into the `technical_improvements` branch,  I'll go ahead and merge now.

PR #351, opened 2023-03-24
Move lumi normalization to topeft processor removing from latter scripts

jlawrenc:
This add the luminosity normalization into the topeft.py processor. The luminosity normalization is also removed from script that use the output of topeft.py. The YieldTools.py script has been tested (via the get_yield_json.py) and has been shown to work. The other two changes in make_cr_and_sr_plots.py and dataDrivenEstimations.py have not been tested.


topcoffea/modules/dataDrivenEstimation.py
kmohrman, line 1: @jlawrenc I think this script might have another part that will need to be edited as well. The `Scale back by 1/lumi all processes but data so they can be used transparently downstream` block rescales the data driven estimation for the fakes by 1/lumi because it is assuming that we will scale it by lumi in the post processing steps (which will no longer happen since we're doing the normalization in the processor). Could you remove that scaling as well?

kmohrman:
@jlawrenc I think the data card maker scripts (make_cards.py or topcoffea/modules/datacard_tools.py) must also scale by the lumi. Could you track down where those happen and remove that scaling as well?

jlawrenc:
Thank @kmohrman for the feedback. I have removed the other lumi normalization that you pointed out. I also tested out the make_cr_and_sr_plots.py and do not see any difference in the plots with the changes.

kmohrman:
Thanks @jlawrenc for these updates. Would you be able to fix/check the following?
* Fix the conflicts with respect to the `technical_improvements` branch 
* Check that the data cards produced by the datacard maker are the same before and after the normalization changes. You can do this by running a command like the following (using the example pkl file in the command) on both the master branch and on this branch:
    ```
    python make_cards.py /afs/crc.nd.edu/user/k/kmohrman/Public/fullR2_files/pkl_files/dec18_fullRun2_withSys_anatest24_np.pkl.gz -C --do-nuisance --var-lst lj0pt ptz -d /some/out/dir --unblind --do-mc-stat
    ```
    And the resulting cards can be compared against the ones here: `/afs/cern.ch/user/k/kmohrman/public/top22006/datacards/at24v01/templates`.

bryates:
@jlawrenc it looks like just flake8 is failing now because the lumi functions are still imported in the files where you removed it (see [CI](https://github.com/TopEFT/topcoffea/actions/runs/4543074334/jobs/8007295369?pr=351#step:4:37))
https://github.com/TopEFT/topcoffea/blob/18f1698943c6ae0ccf41c266191c20ad40184f54/topcoffea/modules/YieldTools.py#L6
https://github.com/TopEFT/topcoffea/blob/18f1698943c6ae0ccf41c266191c20ad40184f54/topcoffea/modules/dataDrivenEstimation.py#L4

bryates:
> Thank @bryates, just for my own edification, this test will fail if something is imported but not used?

Flake8 is a style checker, so it wants to make sure you follow the PEP8 style guide. In this case, importing and not using a module is just a waste of memory, so it marks it is bad.

bryates:
Looks like the new commit introduced a bunch of white spaces it doesn't like, as well as a few imports
https://github.com/TopEFT/topcoffea/actions/runs/4556627573/jobs/8037225783?pr=351#step:4:37

bryates:
> Looks like the new commit introduced a bunch of white spaces it doesn't like, as well as a few imports https://github.com/TopEFT/topcoffea/actions/runs/4556627573/jobs/8037225783?pr=351#step:4:37

Awesome, flake8 now passes! Thanks @jlawrenc

jlawrenc:
All requested tests have been made and all CI checks are passing

kmohrman:
Thanks very much @jlawrenc for these updates and checks. Sorry if I just missed this, but have you been able to verify that the yields in the datacards for a "standard" pkl file (i.e. one that corresponds to a standard analysis run) look consistent as requested [here](https://github.com/TopEFT/topcoffea/pull/351#issuecomment-1486020613)? 

I know we discussed in person that the check does not make sense with our "official" pkl files (since they have not had the lumi normalization applied in the processor), but I think we talked about a check that might make more sense would be to just reproduce an "official" pkl file with the updated processor and run the datacard maker on that.

kmohrman:
Since @jlawrenc has answered all of @bryates's comments and confirmed at the meeting on Friday that he ran the checks described [here](https://github.com/TopEFT/topcoffea/pull/351#issuecomment-1491020119) to confirm that running the datacard maker on the full sized pkl file produced the expected yields, I will go ahead and merge into the `technical_improvements` branch now.

Issue #334, opened 2023-02-28
Apply consistent 4 space indentation across all topcoffea code


kmohrman:
Update all topcoffea code to consistently use 4 space indents, and ensure this will be enforced in the future with an auto-formatter (e.g. `black`).

kmohrman:
Done in PR #352.

kmohrman:
Closed by PR #354

PR #352, opened 2023-03-26
Enforce consistent spacing

kmohrman:
This PR adds [flake8](https://flake8.pycqa.org/en/latest/) to the CI. The PR also applies fixes to many of the topcoffea files in order to fix issues caught by the flake8 test. 

In the flake8 job in the CI yaml file, I added several flake8 error codes to ignore (for formatting things that I don't think we want to bother enforcing). I also entirely excluded a few files from the check (these files are quite old and are close to being deprecated, but are unfortunately still used in a few places, so we probably don't want to remove them yet. But if/when we come to the point where we want to build on these files, we might consider rewriting them instead of building the old ones). Eventually it may be useful to specify the error codes and files to ignore via a flake8 config file (instead of writing them all out in the CI yaml file). 

Addresses Issue #334

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
         uses: py-actions/flake8@v2
         with:
           exclude: "./topcoffea/modules/fileReader.py,./topcoffea/modules/samples.py,./topcoffea/modules/createJSON.py,./topcoffea/modules/WCFit.py"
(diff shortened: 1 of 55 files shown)


kmohrman:
@bryates does this PR look ok to be merged into the `technical_improvements` branch, or would you have any comments?

bryates (approved):

PR #353, opened 2023-03-26
edited condor datacard making command to remove do-nuisance flag

abasnet97:
This PR fixes the issue with the `--do-nuisance` flag not working properly when running datacard maker (make_cards.py) using condor. The issue was that even when the flag was not explicitly called, the resulting datacards would still have the systematics included. It turned out that the datacard making command passed to condor had `--do-nuisance` flag always included. Getting rid of that and adding a new option to use the flag in `run_condor` method resolved the issue.

### DIFF
--- a/analysis/topEFT/make_cards.py
+++ b/analysis/topEFT/make_cards.py
@@ -199,13 +199,6 @@ def main():
     if isinstance(wcs,str):
         wcs = wcs.split(",")
 
-    #if use_condor:
-        # Note:
-        #   The dc in the parent submission is only used to generate the selectedWCs file and figure
-        #   out what channels/samples are available, so we can just drop the systematics to speed
-        #   things up
-        #do_nuis = False
-
     kwargs = {
         "wcs": wcs,
         "lumi_json_path": lumi_json,
--- a/analysis/topEFT/make_cards.py
+++ b/analysis/topEFT/make_cards.py
@@ -56,7 +56,7 @@ source ${USR_DIR}/miniconda3/etc/profile.d/conda.sh
 unset PYTHONPATH
 conda activate ${CONDA_DEFAULT_ENV}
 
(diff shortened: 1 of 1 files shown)


kmohrman:
Thanks very much for fixing this! Just wanted to check that you've verified that these changes did not inadvertently cause any changes when running _without_ condor (i.e. explicitly verified that including or excluding the `--do-nuisance` flag now has the expected effect both without condor and with condor?)

abasnet97:
Yes. I have double-checked that. I ran the script using condor without passing the flag, and I see that the resulting datacard has no systematics included. Again, ran the script using condor with the flag included, and the resulting datcard has systematics included.

analysis/topEFT/make_cards.py
```diff
@@ -197,12 +199,12 @@ def main():
     if isinstance(wcs,str):
         wcs = wcs.split(",")
 
-    if use_condor:
+    #if use_condor:
```
bryates, line 202: You can probably just remove this now

analysis/topEFT/make_cards.py
```diff
@@ -197,12 +199,12 @@ def main():
...
         # Note:
         #   The dc in the parent submission is only used to generate the selectedWCs file and figure
         #   out what channels/samples are available, so we can just drop the systematics to speed
         #   things up
-        do_nuis = False
+        #do_nuis = False
```
bryates, line 207: And this as well

bryates (approved):
Looks good. Just 2 minor comments on commented out code.

analysis/topEFT/make_cards.py
abasnet97: Yes, I forgot about that. Done now!

analysis/topEFT/make_cards.py
abasnet97: Done as well.

kmohrman:
Thanks again for this PR Aashwin. Since it looks like all of the comments have been addressed, I'd be happy to go ahead and merge it now.

PR #354, opened 2023-03-27
Technical improvements

kmohrman:
This branch includes various technical improvements to the topcoffea code (mainly in response to the Issues in the [Technical updates to topcoffea](https://github.com/orgs/TopEFT/projects/1) project). It should not merged yet, as the updates are still in progress.


kmohrman:
Note that we have pinned the `numpy` version in the `environment.yml` file (to numpy=1.23.5) as a workaround to the `numpy.bool` error that we otherwise get (from one of the coffea files).  

This should be revisited at some point (e.g. once coffea has fixed the `numpy.bool` instance, or once conda has figured out how to make a consistent environment that does not lead to this error) to remove the `numpy` version pin. 

See Issue #368.

kmohrman:
The technical_improvements branch has been run through the full workflow. I made the pkl file and datacards on the topcoffea `technical_improvements` branch, and made the workspace on @hnelson's updated branch on the EFTFit repo (PR [here](https://github.com/TopEFT/EFTFit/pull/22)). I also made a CR pkl file and made the CR plots from that as well (just to cross check the normalization) and there were no issues with that. 

The fits (profiled and frozen, unblind) look consistent to me. For the vast majority of the WCs, the differences are less than a percent. But I don't think there's any reason to believe the few O(1%) differences that do exist are due to changes on the `technical_improvements` brach. We don't expect identical results here because I ran with fewer scan points (so that they'd finish faster) so this means the interpolation for the 2sigma number will be different, plus the randomness in the processor can result in very small (sub percent) fluctuations in the ranges.

PR #355, opened 2023-03-28
Combine executors

ywan2:
A new script `run_executor.py` is added in directory `analysis/topEFT` which combines future executor script `run.py` and workqueue executor script `work_queue_run.py`.

- To run WQ, simply type `python run_executor.py ../../topcoffea/cfg/{your file}` since WQ is set as default,
  or `python run_executor.py -x work_queue ../../topcoffea/cfg/{your file}`.
- To run futures, type `python run_executor.py -x futures ../../topcoffea/cfg/{your file}`.


kmohrman:
Thanks very much for adding this @ywan2! It will be really great to have these scripts combined. I will test out the new script soon. 

In the meantime, I'm wondering if you'd be able to make the following updates? 
1. I think we could rename the new script `run_topcoffea.py` (since topcoffea is the processor it runs)
2. Could you fix the `flake8` formatting errors that are causing the CI to fail? 
3. Could you update the Readme in analysis/topEFT to replace the entries for `run.py` and `work_queue_run.py` with an entry for the new `run_topcoffea.py`?
4. Could you also update the `fullR2_run.sh` wrapper script to use the new `run_topcoffea.py` script? 
5. I think we will no longer need the  `run.py` and `work_queue_run.py` scripts, so these can probably be removed.

ywan2:
I made the changes according to @kmohrman comment. CI is apparently still failling but the close parentheses indent should be fixed. Seems like from @bryates, CI is failing because of other issues on the yaml script but not anything related to `run_topcoffea.py` correct?

bryates:
> I made the changes according to @kmohrman comment. CI is apparently still failling but the close parentheses indent should be fixed. Seems like from @bryates, CI is failing because of other issues on the yaml script but not anything related to `run_topcoffea.py` correct?

Yes, sorry, I said yml but it's really in the tests e.g. https://github.com/TopEFT/topcoffea/blob/b87b075daaf00b9c73660b8684f6e1ff8d2b634a/tests/test_futures.py#L10

analysis/topEFT/run_topcoffea.py
```diff
@@ -102,6 +110,13 @@
     # If we don't specify this argument, it will be None, and the processor will fill all hists
     hist_lst = args.hist_list
 
+if executor == "futures":
+    if dotest:
```
kmohrman, line 114: I think this could have unintuitive behavior if the user asks for the wq executor and the `test` option (looks like the code will just silently not actually run in `test` mode).

However, I'm not sure this `dotest` functionality is actually used. I never use it anyway. When I want to run a test, I just explicitly set the number of chunks and/or number of cores myself. Does anyone else (e.g. @bryates) actually use this option? If it's never used, I'd be in favor of just removing it entirely. And I think the `pretend` option is similar. Unless someone actually uses that, I think we should just remove the option.

bryates: Yes, I've used `-t` in the past, but only locally. I agree in principle that it could cause issues on WQ, but I don't know if we'd ever ask WQ to use `-t`, so it should be fine.

kmohrman: Ok @bryates if you use the option I agree we should leave it in. I've added a check to give an error if we try to use that option with anything other than futures. I also added a check to throw an error if someone asks for an executor other than futures or work_queue (since the run script is just set up to run with those two executors).

bryates: Looks good to me.

kmohrman:
@ywan2 I'm sorry for making the bad suggestion to call the run script `run_topcoffea.py`. Since the processor is named topeft (not topcoffea) I think it'd make more sense to call it `run_topeft.py`. So I've pushed that change.

ywan2:
@kmohrman It's all good no worries, and thanks for making the changes. Is there anything else significant that caught you for me to edit?

> @ywan2 I'm sorry for making the bad suggestion to call the run script `run_topcoffea.py`. Since the processor is named topeft (not topcoffea) I think it'd make more sense to call it `run_topeft.py`. So I've pushed that change.

kmohrman:
> @kmohrman It's all good no worries, and thanks for making the changes. Is there anything else significant that caught you for me to edit?
> 
> > @ywan2 I'm sorry for making the bad suggestion to call the run script `run_topcoffea.py`. Since the processor is named topeft (not topcoffea) I think it'd make more sense to call it `run_topeft.py`. So I've pushed that change.

I think it all looks good to me! I'll do a quick check with work queue when I have a chance (probably later this afternoon) and if that looks good I'll be happy to merge at that point, unless @bryates has any further comments.

bryates:
> > @kmohrman It's all good no worries, and thanks for making the changes. Is there anything else significant that caught you for me to edit?
> > > @ywan2 I'm sorry for making the bad suggestion to call the run script `run_topcoffea.py`. Since the processor is named topeft (not topcoffea) I think it'd make more sense to call it `run_topeft.py`. So I've pushed that change.
> 
> I think it all looks good to me! I'll do a quick check with work queue when I have a chance (probably later this afternoon) and if that looks good I'll be happy to merge at that point, unless @bryates has any further comments.

I have nothing else. Go ahead and merge when your checks are done.

PR #356, opened 2023-03-29
Adding training directory to topcoffea/analysis

kmohrman:
This PR adds a new `training` sub directory to the `topcoffea/analysis` directory where we can store simple examples and scripts from tutorials.


kmohrman:
Thanks @abasnet97 for adding the scripts from the 2022 tutorial. @bryates I'm wondering if you could confirm whether or not there are any scripts from the 2021 tutorial (on the `tutorials` branch) that you would like to preserve here (unless they're already included in @abasnet97's commits)? The idea would be to remove the `tutorials` and `topcoffeatutorial2022` after we put anything we want to keep from those branches into the new `analysis/training` directory.

abasnet97:
The notebook I created for 2022 tutorial was based off of Brent's 2021 notebook, so I don't think Brent's repo might have anything different than mine.

bryates:
> The notebook I created for 2022 tutorial was based off of Brent's 2021 notebook, so I don't think Brent's repo might have anything different than mine.

Sorry for the delay. Right, mine is out of date. I did just add the binder link. We should test it once the branch is merged, because I'm pointing it to master which currently does not have this dir strucutre.

bryates (approved):

Issue #339, opened 2023-02-28
Rename variable name for electrons in topeft processor


kmohrman:
It would be good to change the name of the variable used for the electron object to something that is more easily searchable than the current name of `e`. E.g. options may be `ele` or `elec`.

kmohrman:
Closed by PR #354

PR #357, opened 2023-03-29
Change electron name from e to ele in topeft

abasnet97:
In this simple PR, topeft script has been modified to change all instances of `e` to `ele`. 
To verify that all such instances were changed and there were no issues as a result of this, I ran topeft on some signal samples, and there was no issue producing the pkl file.

### DIFF
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -180,7 +180,7 @@ class AnalysisProcessor(processor.ProcessorABC):
 
         # Initialize objects
         met  = events.MET
-        e    = events.Electron
+        ele  = events.Electron
         mu   = events.Muon
         tau  = events.Tau
         jets = events.Jet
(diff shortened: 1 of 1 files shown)


analysis/topEFT/topeft.py
```diff
@@ -252,7 +252,7 @@ def process(self, events):
 
         # Build FO collection
         m_fo = mu[mu.isPres & mu.isLooseM & mu.isFO]
-        e_fo = e[e.isPres & e.isLooseE & e.isFO]
+        e_fo = ele[ele.isPres & ele.isLooseE & ele.isFO]
 
         # Attach the lepton SFs to the electron and muons collections
         AttachElectronSF(e_fo,year=year)
```
bryates, line 258: Looks like there's still a few that are just `e`

bryates: Fair enough. I'm fine with `e_` and `m_`, just wasn't sure if the point of the PR was to replace _all_ `e` with `ele` or just the actual object `e.`.

bryates (approved):

kmohrman:
Thanks very much @abasnet97, I think this will make it much easier when we have to search for the electron object in the code (as searching for `e` usually gives way too many matches..), since @bryates has also approved it, I'll go ahead and merge now.

(Also just to link this to the appropriate issue, this PR fixes Issue #339.)

PR #370, opened 2023-06-11
Plotting script for Lambda plot

kmohrman:
This PR just adds the plotting script that was used to make the Lambda plot of the TOP-22-006 limits. It's by no means a fancy script, but it seems like we might as well have it in the repo.


analysis/topEFT/make_lambda_plot.py
```diff
@@ -0,0 +1,216 @@
...
+    "cQei"   : "$c_{\mathrm{Qe}}^{(\ell)}$"  ,
+    "ctli"   : "$c_{\mathrm{t\ell}}^{(\ell)}$"  ,
+    "ctei"   : "$c_{\mathrm{te}}^{(\ell)}$"  ,
+    "ctlSi"  : "$c_{\mathrm{t}}^{S(\ell)}$" ,
+    "ctlTi"  : "$c_{\mathrm{t}}^{T(\ell)}$" ,
+}
+
+################### From TOP-22-006 ###################
+
+TOP22006_LIMS_DICT = {
```
bryates, line 52: We could think about reading a json file instead of hard coding the values, but that doesn't need to hold up this PR.

bryates (approved):
Thanks for adding this! I'm good with merging now.

analysis/topEFT/make_lambda_plot.py
kmohrman: Agreed

PR #371, opened 2023-07-23
HistEft to scikit hist

btovar:
Main changes are in the histEFT.py file.  API changes:

- Axes need to follow the syntax from schikit HEP.
- For scikit addition to work, axes should be in the same order, have the same name, and have the same label. Thus, HistEFT enforces that there should be exactly one dense axis specified, and that it should be the last one listed in __init__.
- Removed most code related to sumw2. There is still some in YieldTools which I didn't know how to modify.
- Added a small module (axes.py) which has a dictionary with the description of the axes for topeft.py It lists the regular or variable edges as needed.  Currently only variable edges work, as the regular involve a rebinning that can't be done right now.
- Since values() returns the bin counts in hist.Hist, added a new function call  h.eval(values) that evaluates the histogram.
- h.eval has a required argument. If evaluating for sm, use None or {}. It receives either a dict of coefficients or an array.
- Removed the set_wc_coefficients* functions and the internal self_wcs. Instead:
- h.eval(wcs) method returns what `h.set_wc_coefficients(wcs);  h.values()`  would do.
- h.eval has a parameter to include the under/overflow bins. This seems to match what values() did, but it didn't make much sense to me. Should eval only return proper bins?
- Regarding overflow, hist.Hist has only the option to include it or not (boolean). There is not a special option for NaN. NaN gets added to overflow bins.
- The new HistEFT cannot be used as a regular hist.Hist, as If exactly one dense axis needs to be defined. If no wc names coefficients are specified, it is assumed that the wc coefficients for filling the histogram are 1.
- Previous hist iterated on the keys of h._sumw.  Added a function  h.spare_keys() that generates all the possible combination of keys.
- Added some functions for compatibility, such as group() and integrate().
- eft_helper.calc_eft_weights  seems to depend strongly on the structure of these particular histograms, so maybe is a good idea to move it to histEFT.HistEFT
- Some pickled data (e.g. data/triggerSF/*) depend on the old HistEFT module and need to be regenerated before HistEFT can be removed.

All but two unit tests are working:
- tests/test_HistEFT_add.py::test_split_by_terms.  h.split_by_terms needs to be rewritten.
- tests/test_yields.py::test_compare_yields_after_processor.   The values generated are 0, so I think that there is an error on which bins are being looked at (e.g. the undeflow bins) in YieldTools or datacard related files. I didn't understand the code enough to fix this one.

If you have any comments I can fix those, but probably the two failing tests need someone that understand the related physics.

### DIFF
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -224,7 +224,6 @@ class AnalysisProcessor(processor.ProcessorABC):
             # Check to see if the ordering of WCs for this sample matches what want
             if self._samples[dataset]["WCnames"] != self._wc_names_lst:
                 eft_coeffs = efth.remap_coeffs(self._samples[dataset]["WCnames"], self._wc_names_lst, eft_coeffs)
-        eft_w2_coeffs = efth.calc_w2_coeffs(eft_coeffs,self._dtype) if (self._do_errors and eft_coeffs is not None) else None
         # Initialize the out object
         hout = self.accumulator.identity()
 
@@ -889,7 +888,6 @@ class AnalysisProcessor(processor.ProcessorABC):
                                         # Weights and eft coeffs
                                         weights_flat = weight[all_cuts_mask]
                                         eft_coeffs_cut = eft_coeffs[all_cuts_mask] if eft_coeffs is not None else None
-                                        eft_w2_coeffs_cut = eft_w2_coeffs[all_cuts_mask] if eft_w2_coeffs is not None else None
 
 
                                         # Fill the histos
@@ -901,7 +899,6 @@ class AnalysisProcessor(processor.ProcessorABC):
                                             "systematic"    : wgt_fluct,
                                             "weight"        : weights_flat,
                                             "eft_coeff"     : eft_coeffs_cut,
-                                            "eft_err_coeff" : eft_w2_coeffs_cut,
                                         }
 
                                         # Skip histos that are not defined (or not relevant) to given categories
--- a/analysis/topEFT/topeft.py
+++ b/analysis/topEFT/topeft.py
@@ -3,8 +3,10 @@ import copy
 import coffea
 import numpy as np
 import awkward as ak
-np.seterr(divide='ignore', invalid='ignore', over='ignore')
-from coffea import hist, processor
+
+from typing import Sequence, Union
+
+from coffea import processor
 from coffea.util import load
 from coffea.analysis_tools import PackedSelection
 from coffea.lumi_tools import LumiMask
@@ -13,9 +15,14 @@ from topcoffea.modules.GetValuesFromJsons import get_param, get_lumi
 from topcoffea.modules.objects import *
 from topcoffea.modules.corrections import GetBTagSF, ApplyJetCorrections, GetBtagEff, AttachMuonSF, AttachElectronSF, AttachPerLeptonFR, GetPUSF, ApplyRochesterCorrections, ApplyJetSystematics, AttachPSWeights, AttachScaleWeights, GetTriggerSF
 from topcoffea.modules.selection import *
-from topcoffea.modules.HistEFT import HistEFT
 from topcoffea.modules.paths import topcoffea_path
 import topcoffea.modules.eft_helper as efth
+from topcoffea.modules.topeft_axes import info as axes_info
+
+import hist
+from topcoffea.modules.histEFT import HistEFT
+
+np.seterr(divide="ignore", invalid="ignore", over="ignore")
 
 
 # Takes strings as inputs, constructs a string for the full channel name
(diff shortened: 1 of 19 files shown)


topcoffea/modules/eft_helper.py
```diff
@@ -6,7 +6,7 @@
 from numba.typed import List
 import math
 
-@numba.njit
+#@numba.njit
```
bryates, line 9: Does this not work with Numba? How does he performance change without using the JIT?

analysis/topEFT/topeft.py
```diff
@@ -54,26 +58,184 @@ def __init__(self, samples, wc_names_lst=[], hist_lst=None, ecut_threshold=None,
...
-            "ljptsum" : HistEFT("Events", wc_names_lst, hist.Cat("sample", "sample"), hist.Cat("channel", "channel"), hist.Cat("systematic", "Systematic Uncertainty"),hist.Cat("appl", "AR/SR"), hist.Bin("ljptsum", "S$_{T}$ (GeV)", 11, 0, 1100)),
-            "o0pt"    : HistEFT("Events", wc_names_lst, hist.Cat("sample", "sample"), hist.Cat("channel", "channel"), hist.Cat("systematic", "Systematic Uncertainty"),hist.Cat("appl", "AR/SR"), hist.Bin("o0pt",    "Leading l or b jet $p_{T}$ (GeV)", 10, 0, 500)),
-            "bl0pt"   : HistEFT("Events", wc_names_lst, hist.Cat("sample", "sample"), hist.Cat("channel", "channel"), hist.Cat("systematic", "Systematic Uncertainty"),hist.Cat("appl", "AR/SR"), hist.Bin("bl0pt",   "Leading (b+l) $p_{T}$ (GeV)", 10, 0, 500)),
-            "lj0pt"   : HistEFT("Events", wc_names_lst, hist.Cat("sample", "sample"), hist.Cat("channel", "channel"), hist.Cat("systematic", "Systematic Uncertainty"),hist.Cat("appl", "AR/SR"), hist.Bin("lj0pt",   "Leading pt of pair from l+j collection (GeV)", 12, 0, 600)),
-        })
+        self._accumulator = processor.dict_accumulator(
+            {
```
bryates, line 62: This isn't a problem, but I believe coffea+hist now supports dictionaries, but `dict_accumulator` is still fine.

analysis/topEFT/topeft.py
```diff
@@ -54,26 +58,184 @@ def __init__(self, samples, wc_names_lst=[], hist_lst=None, ecut_threshold=None,
...
+                "invmass": HistEFT(
+                    hist.axis.StrCategory([], name="process", growth=True),
```
bryates, line 64: We should have done this a while ago, but it's cleaner to define common axes, like `channel` and `processes` first, and then pass them to the hist calls.

bryates (commented):
Thanks @btovar. I know this is still a WIP, but I left a few comments.

topcoffea/modules/eft_helper.py
btovar: Ah thanks. I disabled it because a previous attempt didn't work, but it should work with the current implementation,

analysis/topEFT/topeft.py
btovar: Sounds good.

tests/test_futures.py
```diff
@@ -13,15 +13,19 @@ def test_topcoffea():
         "topcoffea/json/test_samples/UL17_private_ttH_for_CI.json",
         "-o",
         "output_check_yields",
-        "-p",
-        "analysis/topEFT/histos/"
+        # "-p",
+        # "analysis/topEFT/histos/"
```
bryates, line 17: Was this for local testing? I think we'll need this path restored for `pytest`.

btovar:
Regarding rebining, currently hist does not support it with variable edges.  Thus, there is no easy way to update this line:

https://github.com/TopEFT/topcoffea/blob/501fd5d655783c7862d09bd42f40e2b841acbd0c/topcoffea/modules/datacard_tools.py#L480C1-L480C81

```python
    edge_arr = self.BINNING[km_dist] + [h.axis(km_dist).edges()[-1]]
    h = h.rebin(km_dist,Bin(km_dist,h.axis(km_dist).label,edge_arr))
```

As @kmohrman pointed out, it should be possible to create the histogram with the correct edges to begin with.  I see the edges are created here:

https://github.com/TopEFT/topcoffea/blob/501fd5d655783c7862d09bd42f40e2b841acbd0c/analysis/topEFT/topeft.py#L58

Could I just create those axes with the edge values from the datacard_tools.py module?

klannon:
@bryates @Andrew42 @sscruz Please see the comment from @btovar above.  For some context, it is critically important that we migrate away from coffea histograms to the new scikit-hep based hist class.  Unfortunately, this means at least temporarily migrating away from variable sized rebinning.  I think it's critical that a one or more actual users looks at this and comes up with a solution.  The solution proposed by @btovar and @kmohrman above would certainly work and would have the added advantage of reducing memory usage while filling the histograms, which is often the computational bottleneck.  If you agree that this is a viable solution, can you either help @btovar implement this change in the analysis code or identify someone else who could help?

bryates:
> @bryates @Andrew42 @sscruz Please see the comment from @btovar above. For some context, it is critically important that we migrate away from coffea histograms to the new scikit-hep based hist class. Unfortunately, this means at least temporarily migrating away from variable sized rebinning. I think it's critical that a one or more actual users looks at this and comes up with a solution. The solution proposed by @btovar and @kmohrman above would certainly work and would have the added advantage of reducing memory usage while filling the histograms, which is often the computational bottleneck. If you agree that this is a viable solution, can you either help @btovar implement this change in the analysis code or identify someone else who could help?

@Andrew42 since we eventually hand the histogram off to `uproot` to make the root files, could we use another container that supports variable binning? Maybe we could do this by having the datacard maker convert to `np.histogram` first, and then do the rebinning??

tests/test_HistEFT_add.py
```diff
@@ -1,14 +1,11 @@
...
+import hist
+from topcoffea.modules.histEFT import HistEFT
 import topcoffea.modules.eft_helper as efth
 
-# Let's generate some fake data to use for testing
-nevts = 1000
-rng = np.random.default_rng()
-eft_fit_coeffs = rng.normal(0.3, 0.5, (nevts,276))
-sums = np.sum(eft_fit_coeffs, axis=0)
+import pytest
```
bryates, line 6: I don't think we need to import `pytest`.

btovar:
See #384.

PR #375, opened 2023-08-10
fail test_futures if subprocess fails

btovar:
Previously the test only checked for the presence of an output file. If the file was there from a previous run, the test would be reported as successful even if the python subprocess failed. This pr makes it so that the subprocess call is checked for correct termination.


bryates (approved):
Looks good to me!

kmohrman:
Looks fine to me as well (sorry I did not see this till now).

PR #376, opened 2023-08-17
Initial code for WWZ and some pre-refactoring factorization 

kmohrman:
This PR contains the preliminary code added for the WWZ 4l analysis. The intention is that after the planned `topcoffea` refactoring is complete, all of the WWZ analysis specific code will be extracted into its own GH repository. However, it would be beneficial to merge the changes in this `wwz` branch at this time, as some of the changes anticipate changes we will want to make as part of the refactoring (so it will be good to merge them in order to avoid duplicating work during the refactoring). 

To provide a bit more detail, the majority of the changes are just additions for WWZ (which will be able to be straightforwardly extracted later when we move WWZ to its own GitHub repository), but the main things involving topcoffea and topeft that have changed are these:
* Functions have been extracted from the `analysis/topEFT/make_jsons.py` script and put into a new module in topcoffea/topcoffea/modules 
* The trigger selection in `topcoffea/modules/selection.py` is no longer hard coded to use the trigger dictionaries and dataset dictionaries in that file, instead the dictionaries may be passed as arguments.


Andrew42:
Just took a look over the changed files. It looks like the changes to things unrelated the new WWZ analysis are pretty mild. I'd be fine with this getting merged as is.

PR #378, opened 2023-08-18
Taus pr

jlawrenc:
This PR adds the basics of the hadronic taus to scripts in /topcoffea before the upcoming refactoring.
Main Changes:
-Add Hadronic Tau ID functions to objects.py
-Add Hadronic Tau SFs to corrections.py
-Add new channel setups to selections.py
-Adjust naming convention for Taus in topeft.py
-Add Tau SF files

This adjusts things in the modules not the processor so it will not affect the performance or functionality of the processor since it does not call anything new. The only change to the processor is a change in the naming convention of a tau WP, which does not change the functionality and is not used in the TOP-22-006 analysis. All functions still in development have been marked, but all are functional.


kmohrman:
This PR is currently failing the CI but this is just because one of the root files cannot be downloaded (as it is located at T2 UCSD which is currently down). The PR passed the CI prior to this (and the only change is fixing a merge conflict in the setup folder) so I am ok to merge it at this point so that we can move forward with the refactoring.

PR #380, opened 2023-08-25
Lambda plot formatting

kmohrman:
Updates to the Lambda plot formatting to match what is published in TOP-22-006 supplemental material.


kmohrman:
@bryates can you push your updates to the Lambda formatting (to match the final version the went into the supplemental plots) so that we have it for reference.

bryates (approved):

bryates:
> @bryates can you push your updates to the Lambda formatting (to match the final version the went into the supplemental plots) so that we have it for reference.

I had to actually hack the MPLHEP library in conda to move the text around. I could post instructions on how to do that.

PR #386, opened 2023-10-09
Conda swap space

bryates:
Increase swap space to fixe conda timing out. Same fix that was merged in TopCoffea (https://github.com/TopEFT/topcoffea/pull/11)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,12 @@ jobs:
       - name: Add conda to system path
         run: |
           echo $CONDA/bin >> $GITHUB_PATH 
-
+      
+      - name: Set Swap Space
+        uses: pierotofy/set-swap-space@master
+        with:
+          swap-size-gb: 10
+      
       - name: Create conda env and install dependencies
         run: |
           conda config --set channel_priority strict
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,7 +72,7 @@ jobs:
           echo $CONDA/bin >> $GITHUB_PATH 
       
       - name: Set Swap Space
-        uses: pierotofy/set-swap-space@master
+        uses: pierotofy/set-swap-space@49819abfb41bd9b44fb781159c033dba90353a7c
         with:
           swap-size-gb: 10
       


Andrew42:
@bryates, the only thing that jumps out to me is that, if you can, you should pin the repo that's used to implement this to a specific commit (in this case it would be the most recent one probably). Otherwise, you sort of open yourself up to whatever wild changes might get pushed to a repo you don't control.

bryates:
> @bryates, the only thing that jumps out to me is that, if you can, you should pin the repo that's used to implement this to a specific commit (in this case it would be the most recent one probably). Otherwise, you sort of open yourself up to whatever wild changes might get pushed to a repo you don't control.

This doesn't really use another repo, unless you mean the GitHub action it calls.

klannon:
This seems like a bit of a band-aid.  Why is our CI using more than 7 GB of memory to run?  (You probably already know this but "swap space" is hard drive space that a computer can "swap" memory into if the memory allocated to the process is exhausted.  Of course, switching from memory to hard drive space comes with a huge performance hit which is almost always something that should be avoided unless there is truly no other choice.  Standard Ubuntu GH runners have 7 GB memory, so if we're swapping, it means we're going over that.)

bryates:
> This seems like a bit of a band-aid. Why is our CI using more than 7 GB of memory to run? (You probably already know this but "swap space" is hard drive space that a computer can "swap" memory into if the memory allocated to the process is exhausted. Of course, switching from memory to hard drive space comes with a huge performance hit which is almost always something that should be avoided unless there is truly no other choice. Standard Ubuntu GH runners have 7 GB memory, so if we're swapping, it means we're going over that.)

I'm surprised too, but it seems to be tied to the conda install step. I'm not sure why this recently became an issue, but it seams like our environment is taking up a _lot_ of space as it is downloading and unpacking everything. This isn't an issue with mamba (even though mamba downloads multiple packages at once). My environment on earth (post installation) uses `4.2 GB`, which doesn't line up with my hypothesis.

kmohrman:
I also do not understand why the environment creation started intermittently failing, or why it's due to memory (based on the error message and googling I was guessing some kind of timeout). But since this fix from @bryates  worked and was very non-invasive, we thought it would be a good temporary fix to at least get the CI operational again (before hopefully moving everything to mamba in the future). That was the motivation for merging the same "band-aid" PR for `topcoffea` anyway (and I'm guessing @bryates and @Andrew42 had the same motivations here in `topeft` as well).

Based on the discussion at the meeting last Friday, I think we would still want to move to mamba (and in fact Brent has such a PR open for topcoffea). So I think we still plan to move in that direction (sorry @bryates for being so slow to review the mamba PR, it's on my list but I'd like to try out mamba myself first before we merge that for topcoffea and that has not happened yet.)

klannon:
Thanks for the context.  I had missed that this is happening in the conda step.  Sounds like mamba is the right long-term fix.

bryates:
> Thanks for the context. I had missed that this is happening in the conda step. Sounds like mamba is the right long-term fix.

I agree. I have a PR open now in the topcoffea repo to use mamba (https://github.com/TopEFT/topcoffea/pull/9). It works fine in the CI.

PR #382, opened 2023-08-25
[WIP] Adding fwd regions and lt

kmohrman:
Creating this PR to document the study of forward jet regions, as initial tests showed that this provided increased sensitivity.

PR #387, opened 2023-11-03
Adding fwd regions and lt

bryates:
Replacing PR #382 with updates after repo refactoring.


kmohrman:
Thanks for porting over these changes @bryates. I might suggest holding off on merging this till the HistEFT stuff is in place (just to avoid adding unnecessary complications into the validation). 

Additionally, I would suggest that someone (perhaps Wynona, since it seems to fit in with her general sensitivity improvements studies) would run this and double check the size of the sensitivity improvements before merging. 

Finally, as a general suggestion, I think it would likely be useful to have an easy method of flipping on and off all of these sensitivity improvements from category splitting (e.g. Wynona's off-Z category stuff, and this forward region stuff) since I would imagine that at some point in the future, you will want to know how big of an improvement each change has compared to the "baseline" (if "baseline" is defined as TOP-22-006 categorization). So I think an option to run with the "baseline" categorization would be probably quite useful moving forward.

bryates:
> Thanks for porting over these changes @bryates. I might suggest holding off on merging this till the HistEFT stuff is in place (just to avoid adding unnecessary complications into the validation).
> 
> Additionally, I would suggest that someone (perhaps Wynona, since it seems to fit in with her general sensitivity improvements studies) would run this and double check the size of the sensitivity improvements before merging.
> 
> Finally, as a general suggestion, I think it would likely be useful to have an easy method of flipping on and off all of these sensitivity improvements from category splitting (e.g. Wynona's off-Z category stuff, and this forward region stuff) since I would imagine that at some point in the future, you will want to know how big of an improvement each change has compared to the "baseline" (if "baseline" is defined as TOP-22-006 categorization). So I think an option to run with the "baseline" categorization would be probably quite useful moving forward.

Thanks @kmohrman, I agree with all of this. @ywan2 were you interested in exploring this?

bryates:
This PR was never merged, so I've updated it to work with the current master branch. @Andrew42 and @sscruz should we merge this before #431, or wait until after and fix any conflicts here?

ywan2:
@bryates, it looks like the current version has the default selection including the forward jets, while command line option `--analysis top22006` removes the forward jets. Would you mind doing it the other way? Like having a flag`-f` to turn on forward regions, with the default to be top22006?
This way it's consistent with [ch_update_histEFT](https://github.com/TopEFT/topeft/pull/431/files) branch including the tau and 3l-offZ options.

PR #406, opened 2024-04-01
datacard making update

ywan2:
Two features are added in the datacard making step using `python make_cards.py pkl _np.pkl.gz`:
- We can now set the wcs to certain values while making asimov datacards (excluding `--unblind` option), by including `--wc-vals wc1={number1}, wc2={number2} ...`
An example code to run is:
```
time python make_cards.py {your_file}_np.pkl.gz --do-nuisance --var-lst lj0pt ptz -d {your_datacard_directory} --do-mc-stat --skip-selected-wcs-check --wc-vals ctlTi=0.32, ctlSi=3.1
```
- There is an extra step of systematics check during the binning and zeroing process. When the nominal values are zeros, `FFUp` and `FFDown` are supposed to be zero. If any non-zero value occurs, the systematics is invalid thus an error would be raised. If this error shows up, it would be very likely that the datacards made won't pass text2workspace.py later in combine.


bryates (changes requested):
Thanks @ywan2! This mostly looks good, I just have a few suggestions.

ywan2:
Thank you @bryates for your comments! I made the changes you suggested, except for the last one of "checking non-zero nominal for zero FFUp / FFDown ". There are some channels that have these values during this step but don't raise errors in Combine. Also, I think that keeping "print()" statement might be better than "raise Warning", in case people want to still have the rest of the datacards made to test some things. What do you think?

bryates:
> Thank you @bryates for your comments! I made the changes you suggested, except for the last one of "checking non-zero nominal for zero FFUp / FFDown ". There are some channels that have these values during this step but don't raise errors in Combine. Also, I think that keeping "print()" statement might be better than "raise Warning", in case people want to still have the rest of the datacards made to test some things. What do you think?

@ywan2 Raising a warning should still just print a message and continue. Raising an exception will for sure terminate the script.

bryates:
> > Thank you @bryates for your comments! I made the changes you suggested, except for the last one of "checking non-zero nominal for zero FFUp / FFDown ". There are some channels that have these values during this step but don't raise errors in Combine. Also, I think that keeping "print()" statement might be better than "raise Warning", in case people want to still have the rest of the datacards made to test some things. What do you think?
> 
> @ywan2 Raising a warning should still just print a message and continue. Raising an exception will for sure terminate the script. Are you seeing otherwise?

Ah maybe I was thinking of this (https://stackoverflow.com/questions/3891804/raise-warning-in-python-without-interrupting-program). You can switch back to print statements instead of warnings.

topeft/modules/datacard_tools.py
```diff
@@ -847,22 +847,44 @@ def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins):
...
+                            if syst =="FFUp":
+                                if self.check_zero_arr0 is True and sum(arr[0]) != 0:
+                                    print("Systematics Error arr[0]:Zero values in 'nominal' but non-zero in 'FFUp'")
+                                if self.check_zero_arr1 is True and sum(arr[1]) != 0:
+                                    print("Systematics Error arr[1]:Zero values in 'nominal' but non-zero in 'FFUp'")
+                            if syst == "FFDown":
+                                if self.check_zero_arr0 is True and sum(arr[0]) != 0:
+                                    print("Systematics Error arr[0]:Zero values in 'nominal' but non-zero in 'FFDown'")
+                                if self.check_zero_arr1 is True and sum(arr[1]) != 0:
+                                    print("Systematics Error arr[1]:Zero values in 'nominal' but non-zero in 'FFDown'")
```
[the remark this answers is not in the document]
bryates: @ywan2 just to make sure I understand your comment, there are cases where the nominal has non-zero bins (and hence a non-zero normalization in the txt file) and all zero for a systematic and combine doesn't throw an error? Is this true only for `FF`or others as well?

ywan2:
@bryates, yes I tested the the `raise Warning` statement, it does terminate the code. So I switched back to print statement.

topeft/modules/datacard_tools.py
```diff
@@ -847,22 +847,43 @@ def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins):
...
+                                print("Systematics Error arr[1]:Zero values in 'nominal' but non-zero in 'FFDown'")
 
                         sum_arr = sum(arr[0])
                         if syst == "nominal" and base == "sm":
                             if self.verbose:
                                 print(f"\t{proc_name:<12}: {sum_arr:.4f} {arr[0]}")
                             if not self.use_real_data:
                                 # Create asimov dataset
-                                data_obs += arr
+                                #data_obs += arr
```
bryates, line 882:
```suggestion
```

topeft/modules/datacard_tools.py
```diff
@@ -847,22 +847,43 @@ def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins):
...
                             if arr[1] is not None:
                                 arr[1][negative_bin_mask] = np.zeros_like( arr[1][negative_bin_mask] )  # if there's a sumw2 defined, that one's set to zero as well. Otherwise we will get 0 +/- something, which is compatible with negative
 
-                        syst = sp_key[0]
+                        if len(arr[0][negative_bin_mask]) and syst =="nominal":  # check systematics error for fake factors
+                            if sum(arr[0]) == 0:
+                                check_zero_arr0 = True
+                            if sum(arr[1]) == 0:
+                                check_zero_arr1 = True
+                        if "Up" in syst:
```
bryates, line 865: Thanks @ywan2 for making this change! You could probably remove
`if "Down" in syst`
and change this to
`if "Up" in syst or "Down" in syst`
or even
`if syst != "nominal"`

bryates (changes requested):

topeft/modules/datacard_tools.py
bryates, line 865: @ywan2 after thinking about the terms a bit more, I still think it's important to check _all_ systematics, but with two caveats: 
- The `FF` terms _only_ affect the `fakes_sm` process
- The other terms (e.g. `FSR`, `renorm`, ...) affect every process _but_ `fakes_sm`

Something like this might be the minimal changes needed:
```suggestion
                        check_zero_arr0 = check_zero_arr0 and ('fakes' in proc and 'FF' in syst) or ('fakes' not in proc and 'FF' not in syst)
                        check_zero_arr1 = check_zero_arr1 and ('fakes' in proc and 'FF' in syst) or ('fakes' not in proc and 'FF' not in syst)
                        if "Up" in syst:
```
Now I don't think `proc` is defined here, so see if there's an easy way to pass it or get it here.

Andrew42 (changes requested):
Overall I think these changes look good. I was debating whether or not the best way to pass the WC values is by adding it to the function signature explicitly, e.g. how you had to change all the `def something(arg1,arg2,...)` to `def something(arg1,arg2,...,wcs_dict)`, but i think it's fine. An alternative would have been to just make `wcs_dict` a data member of the `DatacardMaker` class (similar to how `self.drop_syst` or `self.wcs` get specified), but we can leave it as is.

bryates:
Merging this PR now. I've created issues for the remaining items so we can address them later.

PR #411, opened 2024-04-25
Adding separate hist for sumw2

bryates:
This PR addresses https://github.com/TopEFT/topcoffea/issues/27. Instead of altering how `histEFT` works, it is simpler to just have a separate histogram for storing the sum of the square of the weights. This will be part of the `new_histEFT` update, but I wanted a separate PR to make the review easier.

### DIFF
--- a/analysis/topeft_run2/analysis_processor.py
+++ b/analysis/topeft_run2/analysis_processor.py
@@ -67,6 +67,7 @@ class AnalysisProcessor(processor.ProcessorABC):
 
         self._samples = samples
         self._wc_names_lst = wc_names_lst
+        self._isEFT = self._samples[dataset]["WCnames"] != []
         self._dtype = dtype
 
         proc_axis = hist.axis.StrCategory([], name="process", growth=True)
@@ -936,13 +937,16 @@ class AnalysisProcessor(processor.ProcessorABC):
                                         if ((dense_axis_name in ["o0pt","b0pt","bl0pt"]) & ("CR" in ch_name)): continue
 
                                         hout[dense_axis_name].fill(**axes_fill_info_dict)
+                                        sumw2 = np.zeros_like(weights_flat)
+                                        if not self._isEFT: # Only SM (bkg) and data
+                                            sumw2 = np.square(weights_flat)
                                         axes_fill_info_dict = {
                                             dense_axis_name+"_sumw2" : dense_axis_vals[all_cuts_mask],
                                             "channel"       : ch_name,
                                             "appl"          : appl,
                                             "process"       : histAxisName,
                                             "systematic"    : wgt_fluct,
-                                            "weight"        : np.square(weights_flat),
+                                            "weight"        : sumw2,
                                             "eft_coeff"     : eft_coeffs_cut,
                                         }
                                         hout[dense_axis_name+"_sumw2"].fill(**axes_fill_info_dict)
--- a/analysis/topeft_run2/analysis_processor.py
+++ b/analysis/topeft_run2/analysis_processor.py
@@ -937,18 +937,16 @@ class AnalysisProcessor(processor.ProcessorABC):
                                         if ((dense_axis_name in ["o0pt","b0pt","bl0pt"]) & ("CR" in ch_name)): continue
 
                                         hout[dense_axis_name].fill(**axes_fill_info_dict)
-                                        if not isEFT: # Only SM (bkg) and data
-                                            sumw2 = np.square(weights_flat)
-                                            axes_fill_info_dict = {
-                                                dense_axis_name+"_sumw2" : dense_axis_vals[all_cuts_mask],
-                                                "channel"       : ch_name,
-                                                "appl"          : appl,
-                                                "process"       : histAxisName,
-                                                "systematic"    : wgt_fluct,
-                                                "weight"        : sumw2,
-                                                "eft_coeff"     : eft_coeffs_cut,
-                                            }
-                                            hout[dense_axis_name+"_sumw2"].fill(**axes_fill_info_dict)
+                                        axes_fill_info_dict = {
+                                            dense_axis_name+"_sumw2" : dense_axis_vals[all_cuts_mask],
+                                            "channel"       : ch_name,
+                                            "appl"          : appl,
+                                            "process"       : histAxisName,
+                                            "systematic"    : wgt_fluct,
+                                            "weight"        : np.square(weights_flat),
+                                            "eft_coeff"     : eft_coeffs_cut,
+                                        }
+                                        hout[dense_axis_name+"_sumw2"].fill(**axes_fill_info_dict)
 
                                         # Do not loop over lep flavors if not self._split_by_lepton_flavor, it's a waste of time and also we'd fill the hists too many times
                                         if not self._split_by_lepton_flavor: break
--- a/analysis/topeft_run2/analysis_processor.py
+++ b/analysis/topeft_run2/analysis_processor.py
@@ -67,7 +67,6 @@ class AnalysisProcessor(processor.ProcessorABC):
 
         self._samples = samples
         self._wc_names_lst = wc_names_lst
-        self._isEFT = self._samples[dataset]["WCnames"] != []
         self._dtype = dtype
 
         proc_axis = hist.axis.StrCategory([], name="process", growth=True)
@@ -148,6 +147,7 @@ class AnalysisProcessor(processor.ProcessorABC):
 
         # Dataset parameters
         dataset = events.metadata["dataset"]
+        isEFT   = self._samples[dataset]["WCnames"] != []
 
         isData             = self._samples[dataset]["isData"]
         histAxisName       = self._samples[dataset]["histAxisName"]
@@ -938,7 +938,7 @@ class AnalysisProcessor(processor.ProcessorABC):
 
                                         hout[dense_axis_name].fill(**axes_fill_info_dict)
                                         sumw2 = np.zeros_like(weights_flat)
-                                        if not self._isEFT: # Only SM (bkg) and data
+                                        if not isEFT: # Only SM (bkg) and data
                                             sumw2 = np.square(weights_flat)
                                         axes_fill_info_dict = {
                                             dense_axis_name+"_sumw2" : dense_axis_vals[all_cuts_mask],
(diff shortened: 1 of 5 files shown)


topeft/modules/dataDrivenEstimation.py
```diff
@@ -118,7 +118,9 @@ def DDFakes(self):
                         hPromptSub = hPromptSub.remove("systematic", syst_var_idet_rm_lst)
 
                         # now we actually make the subtraction
-                        hPromptSub.scale(-1)
+                        # var(A - B) = var(A) + var(B)
+                        if '_sumw2' not in key:
```
Andrew42, line 122: maybe change this check to be: `key.endswith("_sumw2")` so that it's more explicit that this is checking for a postfix addition.

topeft/modules/datacard_tools.py
```diff
@@ -422,7 +422,7 @@ def read(self,fpath):
         for km_dist, h in self.hists.items():
             if h.empty():
                 continue
-            if self.var_lst and km_dist not in self.var_lst:
+            if self.var_lst and km_dist not in self.var_lst and km_dist.replace("_sumw2", "") not in self.var_lst:
```
Andrew42, line 425: I'm not a big fan of having a boolean check do multiple things in order to perform the actual comparison you want. Specifically in this case I would prefer calling `km_dist.replace('_sumw2',''')` before the `if` clause and save it to a reasonably named temporary variable, which is then used in `if` block.

topeft/modules/datacard_tools.py
```diff
@@ -787,7 +787,14 @@ def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins):
         outf_root_name = self.FNAME_TEMPLATE.format(cat=ch,kmvar=km_dist,ext="root")
 
         h = self.hists[km_dist]
+        try:
+            h_sumw2 = self.hists[km_dist+"_sumw2"]
+        except KeyError:
+            msg = "No sumw2 histogram found! Setting errors to 0"
+            print(msg)
+            h_sumw2 = None
```
Andrew42, line 795: why a `try/except` clause here instead of an `if`? I think this could be rewritten like so:
```
h_sumw2 = None
if f"{km_dist}_sumw2" in self.hists:
  h_sumw2
```
and if you still wanted to have the warning message printed, you could add an `else` block for that situation. If you wanted to be _really_ compact you could rewrite it in one line as:
```
h_sumw2 = self.hists[km_dist+"_sumw2"] if f"{km_dist}_sumw2" in self.hists else None
```
But I'm not the biggest fan of using the ternary conditional in code, unless it's for _really_ simple checks and I think this particular case is borderline too complex

analysis/topeft_run2/analysis_processor.py
```diff
@@ -920,6 +936,16 @@ def process(self, events):
                                         if ((dense_axis_name in ["o0pt","b0pt","bl0pt"]) & ("CR" in ch_name)): continue
 
                                         hout[dense_axis_name].fill(**axes_fill_info_dict)
+                                        axes_fill_info_dict = {
+                                            dense_axis_name+"_sumw2" : dense_axis_vals[all_cuts_mask],
+                                            "channel"       : ch_name,
+                                            "appl"          : appl,
+                                            "process"       : histAxisName,
+                                            "systematic"    : wgt_fluct,
+                                            "weight"        : np.square(weights_flat),
```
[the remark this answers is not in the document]
bryates: Right, it was easier to generalize, and let the datacard maker throw out the signal errors.

topeft/modules/datacard_tools.py
```diff
@@ -787,7 +787,14 @@ def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins):
...
         ch_hist = h.integrate("channel",[ch])
+        ch_sumw2 = h_sumw2.integrate("channel",[ch]) if h_sumw2 is not None else h_sumw2
```
Andrew42, line 797: For this I wonder if we want to swap the order, i.e. have it instead be:
```
ch_sumw2 = h_sumw2 if h_sumw2 is None else h_sumw2.integrate("channel",[ch])
```
I'm sort of ambivalent on this, so I'm fine with keeping it as is. I just find it easier to mentally parse conditionals when there's as few negations as possible.

topeft/modules/datacard_tools.py
```diff
@@ -805,9 +812,10 @@ def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins):
                 if 'fakes' in p and '4l' in ch:
                     continue
                 proc_hist = ch_hist.integrate("process",[p])
+                proc_sumw2 = ch_sumw2.integrate("process",[p]) if ch_sumw2 is not None else ch_sumw2
```
Andrew42, line 815: Same comment as the one posted for L797

topeft/modules/dataDrivenEstimation.py
bryates: Done

topeft/modules/datacard_tools.py
bryates: Done.

topeft/modules/datacard_tools.py
bryates: Ok, I've switched to an `if/else` blocok.

topeft/modules/datacard_tools.py
Andrew42: I also like my suggested swap b/c (to me at least) the line reads more naturally. You make it clear that we leave `h_sumw2` "as is" if it's missing and if it is present do the usual integration over the "channel" axis that's happening in this section of the code.

topeft/modules/datacard_tools.py
bryates: Done

topeft/modules/datacard_tools.py
bryates: Done

analysis/topeft_run2/analysis_processor.py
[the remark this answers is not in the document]
bryates: Alternatively, we could check of the `self._wc_names_lst` is `None` and only fill the weights then.

topeft/modules/datacard_tools.py
```diff
@@ -787,7 +788,15 @@ def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins):
         outf_root_name = self.FNAME_TEMPLATE.format(cat=ch,kmvar=km_dist,ext="root")
 
         h = self.hists[km_dist]
+        h_sumw2 = None
+        if f"{km_dist}_sumw2" in self.hists:
+            h_sumw2 = self.hists[km_dist+"_sumw2"]
+        else:
+            msg = "No sumw2 histogram found! Setting errors to 0"
+            print(msg)
+            h_sumw2 = None
```
Andrew42, line 797: This last line `h_sumw2 = None` is redundant, but doesn't hurt leaving it in. So fine to keep it or drop it.

topeft/modules/datacard_tools.py
```diff
@@ -422,7 +422,8 @@ def read(self,fpath):
         for km_dist, h in self.hists.items():
             if h.empty():
                 continue
-            if self.var_lst and km_dist not in self.var_lst:
+            km_dist_sumw = km_dist.replace('_sumw2','')
+            if self.var_lst and km_dist not in self.var_lst and km_dist_sumw not in self.var_lst:
```
Andrew42, line 426: So now with the restructured `if` check, I'm wondering what this last part of the boolean check is actually looking for? I read it as: 1st we want to make sure that `self.var_lst` is non-empty, 2nd we want to make sure that `km_dist` is not an element of that list, but then I'm not sure I understand what the 3rd check is looking for. It sounds to me like `self.hists,items()` will contain both the strings `"some_dist"` and `"some_dist_sumw2"`, which I think is fine, but it gives me pause when reading this line, since it means we're splitting how the `for` loop works depending on if a certain distribution exists _and_ if it also has this new `sumw2` histogram.

What you have here might make total sense and we can leave it as is, but maybe you can try to explain the thought process here to help me understand what we're checking for.

bryates: What I'm doing here is only letting it load the original `km_dist` _and_ the `km_dist_suw2`. Everything else is skipped.

topeft/modules/datacard_tools.py
bryates: Removed

topeft/modules/datacard_tools.py
Andrew42: I wonder if it would make it easier to read and more self documenting to have something like:
```
km_dist_missing = km_dist not in self.var_lst
km_base_dist_missing = km_dist.replace('_sumw2','')
if self.var_lst and km_dist_missing and km_base_dist_missing:
```
But I'd still like to hear you're interpretation of what this check is supposed to be doing

Andrew42: I'm not sure I understand what you mean about letting it load _only_ the original and the `sumw2` version. Wouldn't any "original" distribution that shows up in `self.var_lst` pass this check regardless if it has a `sumw2` complement?

Andrew42: I'd maybe describe this check instead as "We only want to include a `sumw2` histogram if it's corresponding "base" `km_dist` is part of the `self.var_lst` list". Is that a reasonable interpretation?

bryates: Yes, the original distribution always passes, but the `sumw2` histogram doesn't pass by default because e.g. `lj0pt_sumw2` _isn't_ in `self.var_lst`

bryates: > I wonder if it would make it easier to read and more self documenting to have something like:
> 
> ```
> km_dist_missing = km_dist not in self.var_lst
> km_base_dist_missing = km_dist.replace('_sumw2','') not in self.var_lst
> if self.var_lst and km_dist_missing and km_base_dist_missing:
> ```
> 
> But I'd still like to hear you're interpretation of what this check is supposed to be doing (also I realize I'm being a hypocrite regarding my earlier comment about not having `bool` checks do multiple things at the same time here)

I like this suggestion, it is a bit easier to read.

Andrew42:
Ok, after the changes you made to my comments, I think I'm good with having this PR merged.

analysis/topeft_run2/analysis_processor.py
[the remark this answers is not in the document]
Andrew42: I like the idea of having a check on the `self._wc_names_lst` as it makes it more explicit that these `sumw2` histograms should only exist for the non-EFT samples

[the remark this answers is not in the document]
bryates: I just remembered the whole point of this is we _need_ the MC errors from the SM point of our private samples, since that error goes into `fakes_sm`. I've restored it back to my original implementation.

PR #440, opened 2024-11-08
Use run_dict from TopCoffea

bryates:
This PR updates `corrections.py` to use the new run_dict json file in TopCoffea.


bryates:
@Andrew42 @sscruz any comments on this PR?

PR #444, opened 2024-11-21
Adding fwd regions and lt

bryates:
Replacing #387, targeting `ch_update_histeft`.


ywan2:
@bryates Thanks for remaking the branch based on the `ch_update_histeft`. The analysis processor runs fine with and without the new flag, so it's good!
However, two things:
1. I spotted that you put the new channels in `channels/ch_lst.json` file. We actually are using `channels/ch_lst_test.json` in the analysis processor (sorry it's confusing), so it would be more convenient to move things there instead. 
2. Since there are new channels, would you also want to edit the `datacard_post_processing.py` to include the selection too? Also in the similar format like our offZ `-z`, and tau `-t`?

bryates:
> @bryates Thanks for remaking the branch based on the `ch_update_histeft`. The analysis processor runs fine with and without the new flag, so it's good! However, since there are new channels, would you also want to edit the `datacard_post_processing.py` to include the selection too? Also in the similar format like our offZ `-z`, and tau `-t`?

Thanks @ywan2, I've added some stuff to `datacard_post_processing.py`. Let me know if I've missed anything.

PR #449, opened 2024-12-04
New jsons for including extensions in the same jsons as the original sample, and updated

anpicci:
This PR includes updated jsons from @anehnis and updated xsec for the TTToLNu2Q sample


bryates (approved):
Thanks @anpicci, this looks good to me. @Andrew42 and @sscruz any comments?

input_samples/sample_jsons/background_samples/central_2022/CheckJsons.py
```diff
@@ -0,0 +1,39 @@
...
+    current_directory = os.getcwd()
+    results = check_files_in_json(current_directory, BASE_PATH)
+
+    for json_file, result in results.items():
+        print(f"JSON File: {json_file}")
+        print(f"  Total Files: {result['total_files']}")
+        print(f"  Missing Files: {result['missing_count']}")
+        if result['missing_files']:
+            print(f"  Missing Files Sample: {result['missing_files'][:5]}")
+        print()
```
bryates, line 39: Is this empty print just to make a new line?

input_samples/sample_jsons/background_samples/central_2022/FixJsons.py
```diff
@@ -0,0 +1,37 @@
+import os
```
bryates, line 1: I'm curious what this file is for. Did some of the central samples go missing after we made the json files? If so, does this properly fix the sum of weights, or should we rerun `make_jsons.py` instead?

input_samples/sample_jsons/background_samples/central_2022/CheckJsons.py
anpicci: yes

bryates: You could add `\n` to the previous string, but I'm fine either way.

anpicci:
@bryates I think we merge this PR

PR #450, opened 2024-12-04
Object selection organization and run 3D ID

sscruz:


topeft/modules/corrections.py
```diff
@@ -1168,33 +1168,7 @@ def AttachElectronSF(electrons, year, looseWP=None):
...
-            ceval[egm_tag].evaluate(egm_year, "sfup", looseWP, *egm_args)
-        )
-        loose_do_flat = ak.where(
-            ~pt_mask,
-            1,
-            ceval[egm_tag].evaluate(egm_year, "sfdown", looseWP, *egm_args)
-        )
-        loose_sf = ak.unflatten(loose_sf_flat, ak.num(pt))
-        loose_up = ak.unflatten(loose_up_flat, ak.num(pt))
-        loose_do = ak.unflatten(loose_do_flat, ak.num(pt))
```
bryates, line 1196: Is this code we need to keep?

anpicci: @bryates if we don't plan to use any loose flag from EGM POG (if I remember correctly what @sscruz reported in the previous PR), I am fine with removing this part for the Run 3 analysis

bryates: I wasn't sure if setting the loose flag to `None` was a temporary thing until loose SFs are ready. If we don't expect to have loose SFs for Run 3, then we can remove these lines. Otherwise, we should put them back in and maybe raise an exception until they're ready.

anpicci: Got it. So, according to this comment from @sscruz :
https://github.com/TopEFT/topeft/pull/445#discussion_r1867310971

I would say, we can add to the original if statement the condition:
```
looseWP is not None
```
and it should properly handle the scenario where this if section is not deleted, but not actually used until we decide to implement any SFs for the EGM POG MVA, if we will eventually use it. What do you think?

bryates: Sure, that works for me

anpicci: @sscruz let me know if you want me to apply this modification

analysis/topeft_run2/analysis_processor.py
```diff
@@ -322,7 +314,7 @@ def process(self, events):
         e_fo = ele[ele.isPres & ele.isLooseE & ele.isFO]
 
         # Attach the lepton SFs to the electron and muons collections
-        AttachElectronSF(e_fo, year=year, looseWP=ele_pres_id_tag) #Run3 ready
+        AttachElectronSF(e_fo, year=year, looseWP="none" if is_run3 else "wpLnoiso") #Run3 ready
```
bryates, line 317:
```suggestion
        AttachElectronSF(e_fo, year=year, looseWP=None if is_run3 else "wpLnoiso") #Run3 ready
```

sscruz: I'm confused by this change: shouldn't this force the code to crash in run3 ( https://github.com/TopEFT/topeft/pull/450/files#diff-bad2815db1cdd28a50951857cadd189aad367df4d391bd9fc43a6d7187715018L1067  ) ?

bryates: Yeah, sorry. Ignore this comment.

anpicci (approved):

PR #452, opened 2024-12-16
3l fwd

bryates:
Adding new 3l forward jet regions, as talked about in the last meeting. Most of this will isolate the extra parton in the ttW process, but the new 1 jet forward region is designed to isolate the single-top spectator quark.


analysis/topeft_run2/analysis_processor.py
```diff
@@ -721,6 +721,10 @@ def process(self, events):
                 preselections.add("2lss_fwd", (events.is2l & pass_trg & fwdjet_mask))
                 preselections.add("2l_fwd_p", (chargel0_p & fwdjet_mask))
                 preselections.add("2l_fwd_m", (chargel0_m & fwdjet_mask))
+                preselections.add("3l_fwd", (events.is3l & pass_trg))
```
ywan2, line 724: @bryates This is perhaps a minor thing. I see that you include the `fwdjet_mask` in `ch_lst_test.json` instead of in the `3l_fwd` mask in the processor. But 2lss_fwd `fwdjet_mask` does the other way. Do we want to keep the selection codes for 2lss_fwd and 3l_fwd more consistent just to be easier for others to read/understand?

bryates: Great catch. Let's move it to the json file for `2lss`, that way keep the processor a bit more generalized.

analysis/btagMCeff/btagMCeff.py
```diff
@@ -32,15 +32,15 @@ def __init__(self, samples):
         jeta_axis = hist.axis.Regular(25, -2.5, 2.5, name="eta", label=r"Jet \eta (GeV)")
         jaeta_axis = hist.axis.Variable([0, 1, 1.8, 2.4], name="abseta", label=r"Jet \eta (GeV)")
         Flav_axis = hist.axis.StrCategory([], name="Flav", growth=True)
-        flav_axis = hist.axis.IntCategory([], name="flav", growth=True)
+        flav_axis = hist.axis.Regular(5, 0, 5, name="flav", label="jet flav")
```
bryates, line 35: @ywan2 did you push these change by accident?

ywan2:
> @ywan2 was this intended? [8ebdfb1#diff-c51d434110b01b664b71f73a6904805333c86530877d042306c36580b00ebd73L724](https://github.com/TopEFT/topeft/commit/8ebdfb12e31b3bf1c4ca86efa8c420d4b7800034#diff-c51d434110b01b664b71f73a6904805333c86530877d042306c36580b00ebd73L724)

If you check the `ch_lst_test.json` file [here](https://github.com/TopEFT/topeft/blob/8ebdfb12e31b3bf1c4ca86efa8c420d4b7800034/topeft/channels/ch_lst_test.json#L640-L658), you can see that these masks are actually redundant and can be removed.

bryates:
> > @ywan2 was this intended? [8ebdfb1#diff-c51d434110b01b664b71f73a6904805333c86530877d042306c36580b00ebd73L724](https://github.com/TopEFT/topeft/commit/8ebdfb12e31b3bf1c4ca86efa8c420d4b7800034#diff-c51d434110b01b664b71f73a6904805333c86530877d042306c36580b00ebd73L724)
> 
> If you check the `ch_lst_test.json` file [here](https://github.com/TopEFT/topeft/blob/8ebdfb12e31b3bf1c4ca86efa8c420d4b7800034/topeft/channels/ch_lst_test.json#L640-L658), you can see that these masks are actually redundant and can be removed.

Ok thanks for confirming.

analysis/btagMCeff/btagMCeff.py
ywan2: These I pushed by accident. Sorry about it! Although it does fix the script to run, but I think I will make these changes in another branch to avoid confusion.

bryates: Thanks! Sounds like a good plan.

PR #457, opened 2025-02-24
added jsons for EFT TWZToLL samples

abasnet97:
This simple PR adds the jsons for the EFT tWZtoLL samples obtained from Alberto.

**For future reference**
When making the jsons, a couple of tweaks had to be made. 

1. The nAOD didn't have the built-in `LHEReweightingWeight` string to extract the `WCnames`, so we manually put in the `WCnames` by getting relevant information from Alberto. 
2. The `PSWeight` attribute of these nAODs didn't have the right mapping that our script (`topcoffea/modules/corrections.py`) expects, so we had to get the correct mapping information from Alberto and had to hack the `AddPSWeights()` function in the script to store the ISR/FSR variations directly instead of creating a map. It is safe to do so since we are just dealing with a single sample set. This is achieved in the following way:
```
    #Add up variation event weights
    events['ISRUp'] = events.PSWeight[:, 8]
    events['FSRUp'] = events.PSWeight[:, 9]

    #Add down variation event weights
    events['ISRDown'] = events.PSWeight[:, 6]
    events['FSRDown'] = events.PSWeight[:, 7]


bryates (approved):
THanks @abasnet97, this looks good. I was considering asking you to add logic to handle the weights in the future, but as this is a special case (where the EFT structure constants are also in the `PSweights` field) I don't think it's worth committing the extra code.

abasnet97:
> THanks @abasnet97, this looks good. I was considering asking you to add logic to handle the weights in the future, but as this is a special case (where the EFT structure constants are also in the `PSweights` field) I don't think it's worth committing the extra code.

Hi @bryates, yes I did consider that, but like you said, this is a very special case, and I don't know if we will be encountering this much in the future. I guess the important thing is to keep in mind that not everyone handles these weights in the same manner as we do, and if we are obtaining samples from other groups, we should be open to modifying our codes if need be to handle them. But again, it will probably be on a case-by-case basis.

abasnet97:
@Andrew42 do you have any comments on this?

Andrew42:
Since you couldn't get the WC name info from the nAOD, how did you get the sum of weights? Do these samples have proper EFT weight information? e.g. did the skimming step properly compute the quadratic parameterizations for each event in the samples?

abasnet97:
> Since you couldn't get the WC name info from the nAOD, how did you get the sum of weights? Do these samples have proper EFT weight information? e.g. did the skimming step properly compute the quadratic parameterizations for each event in the samples?

@Andrew42 that's a very good point, and I completely overlooked this. 
I did just realize that this sample's `events` collection does not have `EFTfitCoefficients` field (or something similar).

PR #458, opened 2025-02-25
Handle SMEFTsim topU3l <-> dim6top rotations

bryates:
This is an attempt to handle rotations between SMEFTsim topU3l and dim6top in IM. The changes to the datacard maker could use some refactoring, but everything works.


bryates:
I should add that the `text2workspace` step crashes on glados when using `expr` for WCs, but it runs on lxplus. We will need to debug this in the future.

Andrew42:
Interesting, how does this deal with the decomposition step? i.e. where we break up the parameterization into separate "sub-"processes in order to encode the WC dependence.

bryates:
> Interesting, how does this deal with the decomposition step? i.e. where we break up the parameterization into separate "sub-"processes in order to encode the WC dependence.

We don't do that anymore with IM, everything is just stored in the scaling file. If we wanted to do something similar in AAC, it would probably still be casting the "new" WCs as functions of the "old".

Andrew42:
> > Interesting, how does this deal with the decomposition step? i.e. where we break up the parameterization into separate "sub-"processes in order to encode the WC dependence.
> 
> We don't do that anymore with IM, everything is just stored in the scaling file. If we wanted to do something similar in AAC, it would probably still be casting the "new" WCs as functions of the "old".

Ah that's right, sorry! That does make things a lot simpler.

Looking at the code, I'm trying to understand what's going on in the block where you open/load the yaml file. My naive understanding is that the yaml file is mean to basically define a set of equations/functions whose arguments are purely WCs of the other basis, so when combine does the minimization in terms of those WCs the effect gets translated into the appropriate scaling in terms of the new/other WCs. What I don't get is all this manipulation you then need to do after you've loaded the yaml file. I also don't understand the use of `eval` in this block of code.

topeft/params/SMEFTsim-topU3l_dim6top.yml
```diff
@@ -0,0 +1,23 @@
...
+
+"cQQ1": ["expr::SEMFTsim_cQQ1('0.5 * @0', {cQQ1})", ["cQQ1"]]
+"ctt": "ctt1"
+
+# ctG relative minus sign
+"ctGRe": ["expr::ctGRe('-1 * @0', {ctG})", ["ctG"]]
+"ctGIm": ["expr::ctGIm('-1 * @0', {ctG})", ["ctG"]]
+
+# ctW,ctB <-> ctW,ctZ
+"ctWRe": ["expr::ctWRe('-1 * @0', {ctW}", ["ctW"]]
```
Andrew42, line 21: I think this line is missing a closing parenthesis ")" after the `{ctW}`

bryates:
> > > Interesting, how does this deal with the decomposition step? i.e. where we break up the parameterization into separate "sub-"processes in order to encode the WC dependence.
> > 
> > 
> > We don't do that anymore with IM, everything is just stored in the scaling file. If we wanted to do something similar in AAC, it would probably still be casting the "new" WCs as functions of the "old".
> 
> Ah that's right, sorry! That does make things a lot simpler.
> 
> Looking at the code, I'm trying to understand what's going on in the block where you open/load the yaml file. My naive understanding is that the yaml file is mean to basically define a set of equations/functions whose arguments are purely WCs of the other basis, so when combine does the minimization in terms of those WCs the effect gets translated into the appropriate scaling in terms of the new/other WCs. What I don't get is all this manipulation you then need to do after you've loaded the yaml file. I also don't understand the use of `eval` in this block of code.

The manipulations were my (probably inefficient) way to keep things general. I saved the cosine of the Weinberg angle as `CW` instead of hard coding it. The `eval` call is to compute `SW`.

topeft/params/SMEFTsim-topU3l_dim6top.yml
bryates: Good catch

topeft/modules/datacard_tools.py
```diff
@@ -306,6 +307,26 @@ def __init__(self,pkl_path,**kwargs):
...
+                        if "CW" in vv or "SW" in vv:
+                            v[i] = vv.replace("{CW}", str(self.rotate["CW"])).replace("{SW}", str(self.rotate["SW"]))
+                    wcs = v[1]
+                    v   = v[0]
+                    wcs = {wc: wc+"[0, " + str(self.wc_ranges[wc][0]) + ", " + str(self.wc_ranges[wc][1]) + "]" for wc in wcs}
+                    v = v.format(**wcs)
+                    self.rotate[k] = v
+                if isinstance(v, str) and v in self.wc_ranges:
+                    v = v.replace(v, v+"[0, " + str(self.wc_ranges[v][0]) + ", " + str(self.wc_ranges[v][1]) + "]")
+                    self.rotate[k] = v
```
Andrew42, line 328: This entire block does not need to be nested under the `with` block

bryates: Good point

Andrew42: The `if "CW" in v or "SW" in v` check is misleading, since the entire point of this block is to essentially evaluate the string `"np.sqrt(1 - np.square({CW}))"`. I guess it allows you to define the function for `CW` if you wanted to change it in the future, but that seems unnecessary to me?

bryates: Yeah I was trying to keep it fancy but it wound up being a headache. We can probably just store the numerical values. I was thinking CW might change with other WCs, but it doesn't, it only gets multiplied by WCs (that we don't currently use).

Andrew42: As per my continued war on python list comprehensions, I would propose to change this line:
```python
wcs = v[1]
v   = v[0]
wcs = {wc: wc+"[0, " + str(self.wc_ranges[wc][0]) + ", " + str(self.wc_ranges[wc][1]) + "]" for wc in wcs}
```
to the (imo) more readable
```python
wcs = {}
for wc in v[1]:
    lo,hi = self.wc_ranges[wc]
    wcs[wc] = f"{wc}[0,{lo},{hi}]"
v = v[0]
```

bryates: Fair enough. I prefer comprehensions for speed, but this case is probably small enough that it doesn't matter.

topeft/modules/datacard_tools.py
```diff
@@ -306,6 +307,28 @@ def __init__(self,pkl_path,**kwargs):
...
+                    if "CW" in vv or "SW" in vv:
+                        v[i] = vv.replace("{CW}", str(self.rotate["CW"])).replace("{SW}", str(self.rotate["SW"]))
+                wcs = {}
+                for wc in v[1]:
+                    lo,hi = self.wc_ranges[wc]
+                    wcs[wc] = f"{wc}[0,{lo},{hi}]"
+                v = v[0]
+                self.rotate[k] = v
+            if isinstance(v, str) and v in self.wc_ranges:
+                v = v.replace(v, v+"[0, " + str(self.wc_ranges[v][0]) + ", " + str(self.wc_ranges[v][1]) + "]")
```
Andrew42, line 329: sorry for being a both, but could we also replace this line:
```python
v = v.replace(v, v+"[0, " + str(self.wc_ranges[v][0]) + ", " + str(self.wc_ranges[v][1]) + "]")
```
with
```python
lo,hi = self.wc_ranges[v]
v = v.replace(v,f"{v}[0,{lo},{hi}]")
```
It makes it a lot easier for my brain to parse what's going on here

bryates: Done

Andrew42: thanks

topeft/modules/datacard_tools.py
```diff
@@ -306,6 +307,29 @@ def __init__(self,pkl_path,**kwargs):
...
+            elif isinstance(v, list):
+                for i,vv in enumerate(v):
+                    if "CW" in vv or "SW" in vv:
+                        v[i] = vv.replace("{CW}", str(self.rotate["CW"])).replace("{SW}", str(self.rotate["SW"]))
+                wcs = {}
+                for wc in v[1]:
+                    lo,hi = self.wc_ranges[wc]
+                    wcs[wc] = f"{wc}[0,{lo},{hi}]"
+                v = v[0]
+                self.rotate[k] = v
```
Andrew42, line 327: I think you accidentally removed the `v = v.format(**wcs)` line, which is still needed. So it should be:
```python
wcs = {}
for wc in v[1]:
    lo,hi = self.wc_ranges[wc]
    wcs[wc] = f"{wc}[0,{lo},{hi}]"
v = v[0]
v = v.format(**wcs)
self.rotate[k] = v
```

bryates: Woops, should be fixed now

topeft/params/SMEFTsim-topU3l_dim6top.yml
```diff
@@ -0,0 +1,23 @@
...
+"cQj11": "cQq11"
+"cQj18": "cQq81"
+"cQj38": "cQq83"
+
+"cQQ1": ["expr::SEMFTsim_cQQ1('0.5 * @0', {cQQ1})", ["cQQ1"]]
+"ctt": "ctt1"
+
+# ctG relative minus sign
+"ctGRe": ["expr::ctGRe('-1 * @0', {ctG})", ["ctG"]]
+"ctGIm": ["expr::ctGIm('-1 * @0', {ctG})", ["ctG"]]
```
Andrew42, line 18: Does this mean that `ctGRe` and `ctGIm` will be scaled/set/controlled by a single parameter (e.g. 'ctG')? Do we not assume the imaginary part for this (and other) operators is 0?

bryates: Oh that was a mistake, it should be ctGI for ctGIm

topeft/params/SMEFTsim-topU3l_dim6top.yml
```diff
@@ -0,0 +1,23 @@
...
+"ctt": "ctt1"
+
+# ctG relative minus sign
+"ctGRe": ["expr::ctGRe('-1 * @0', {ctG})", ["ctG"]]
+"ctGIm": ["expr::ctGIm('-1 * @0', {ctGI})", ["ctGI"]]
+
+# ctW,ctB <-> ctW,ctZ
+"ctWRe": ["expr::ctWRe('-1 * @0', {ctW})", ["ctW"]]
+"ctBRe": ["expr::ctBRe('@0 / {SW} - @1 * {CW}/{SW} ', {ctZ}, {ctW})", ["ctZ", "ctW"]]
+"ctBRe": ["expr::ctBRe('@0 / {SW} - @1 * {CW}/{SW} ', {ctZ}, {ctW})", ["ctZ", "ctW"]] # {ctZ}/SW - {ctW}/TW (TW = tan Weinberg angle)
```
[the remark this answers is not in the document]
bryates: Good catch. Maybe it's was due to a merge conflict when I added the comment. Removed now.

topeft/modules/datacard_tools.py
```diff
@@ -306,6 +307,30 @@ def __init__(self,pkl_path,**kwargs):
...
+                for wc in v[1]:
+                    lo,hi = self.wc_ranges[wc]
+                    wcs[wc] = f"{wc}[0,{lo},{hi}]"
+                v = v[0]
+                v = v.format(**wcs)
+                self.rotate[k] = v
+            if isinstance(v, str) and v in self.wc_ranges:
+                lo,hi = self.wc_ranges[v]
+                v = v.replace(v,f"{v}[0,{lo},{hi}]")
+                self.rotate[k] = v
```
Andrew42, line 332: I'd like to point out that this block of code implicitly assumes that the "CW" and "SW" keys get processed before any of the other keys, otherwise the "SW" substitutions done later wouldn't be correct.

Apparently starting with python 3.7 dictionaries are guaranteed to preserve the keys ordering based on the order they were inserted into the dictionary. So this block of code is technically safe, but makes me uneasy based on the general understanding that dictionaries are supposed to be unordered maps.

How do you feel about refactoring it to look something like this:
```python
self.rotate = {}
with open(topeft_path("params/SMEFTsim-topU3l_dim6top.yml")) as f:
    yml = yaml.load(f,Loader=yaml.CLoader)

self.rotate["CW"] = yml.pop("CW")
self.rotate["SW"] = eval(yml.pop("SW").format(CW=self.rotate["CW"]))

for k,v in yml:
    if isinstance(v, list):
        kwargs = {}
        for wc in v[1]:
            lo,hi = self.wc_ranges[wc]
            kwargs[wc] = f"{wc}[0,{lo},{hi}]"
        if "{CW}" in v[0]:
            kwargs["CW"] = self.rotate["CW"]
        if "{SW}" in v[0]:
            kwargs["SW"] = self.rotate["SW"]
        self.rotate[k] = v[0].format(**kwargs)
    if isinstance(v, str) and v in self.wc_ranges:
        lo,hi = self.wc_ranges[v]
        v = v.replace(v,f"{v}[0,{lo},{hi}]")
        self.rotate[k] = v
```
It separates the computation of the CW/SW constants from the computation that parses the new WC mappings, so the main `for` loop doesn't have to deal with as many edge cases. It also replaces one of the inner `for` loops with simple direct `if` statements for the cases where we need to substitute in the values of CW/SW, which I think is a lot easier to understand.

It's still not quite as clean as I would like. The `v[0]` and `v[1]` variables are a bit too opaque and I'm not the biggest fan of having "CW" and "SW" keys be part of the `self.rotate` dictionary when they aren't actually WCs, but I don't think that these are all that serious that they should block us.

bryates: I think this is a good idea. How about an exception if `v` is not in `wc_ranges`?

Andrew42: yeah, I like the explicit check to raise the exception as it makes it more obvious what the exact case is we are wanting to check for. Overall I think the changes look good and I'd be fine with this getting merged.

bryates: Ok sounds good!

PR #465, opened 2025-06-03
Minor fix in datacard functions

ywan2:
`datacard_tools.py` is updated for:
 - new topcoffea scalings function parsing.
 - asimov datacard fix function fix.
 
As for changing the datacard WC values, it's still subject to better changes.

### DIFF
--- a/topeft/modules/datacard_tools.py
+++ b/topeft/modules/datacard_tools.py
@@ -858,7 +858,7 @@ class DatacardMaker():
                         for sp_key,arr in data_sm.items():
                             data_obs += arr
                     if not self.use_real_data and wcs_dict is not None:
-                    # Create asimov dataset, 'vals' are WCs specified at the command-line
+                        # Create asimov dataset, 'vals' are WCs specified at the command-line
                         vals = wcs_dict
                         decomposed_templates = self.decompose(proc_hist,proc_sumw2,wcs,vals)
                         data_sm = decomposed_templates.pop("sm")


topeft/modules/datacard_tools.py
```diff
@@ -963,10 +968,10 @@ def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins, wcs_dict):
                 # obtain the scalings for scalings.json file
                 if p in self.SIGNALS:
                     if self.wc_scalings:
-                        scalings = h[{'channel':ch,'process':p,'systematic':'nominal'}].make_scaling(self.wc_scalings)
+                        scalings = h[{'channel':ch,'process':p,'systematic':'nominal'}].make_scaling(flow='show', wc_list=self.wc_scalings)
```
bryates, line 971: Is this what we want? I believe `flow='show'` includes the overflow bins. Haven't we already clipped them off in the datacard maker?

bryates: Do you get different results without it?

ywan2: I did not try the other way, I could test so too.
Just for reference: https://github.com/TopEFT/topcoffea/pull/59

bryates: Thanks, I forgot the exact meaning. This is correct.

topeft/modules/datacard_tools.py
```diff
@@ -857,6 +857,14 @@ def analyze(self,km_dist,ch,selected_wcs, crop_negative_bins, wcs_dict):
...
                         for sp_key,arr in data_sm.items():
                             data_obs += arr
+                    if not self.use_real_data and wcs_dict is not None:
+                        # Create asimov dataset, 'vals' are WCs specified at the command-line
+                        vals = wcs_dict
+                        decomposed_templates = self.decompose(proc_hist,proc_sumw2,wcs,vals)
+                        data_sm = decomposed_templates.pop("sm")
+                        for sp_key,arr in data_sm.items():
+                            if sp_key.systematic == "nominal":
+                                data_obs += arr
```
bryates, line 867: Just double checking this now gives the correct Asimov observations? If we use `-t -1` in combine we see no change in the limits?

ywan2: This section is the part that targets the non SM case where some WCs get set to a different values (but I should add a block to say this is subject to changes). What I did was I remove this section from later lines, and that fix the asimov data issue. Now if we were to look at SM datacards, it should be the same as `-t -1`.

bryates (approved):

PR #492, opened 2025-09-03
Editing plotting script to include stat uncertainties

abasnet97:
This PR edits the `make_cr_and_sr_plots.py` to correctly handle the statistical uncertainties in the plot. In the existing code, the stat err is calculated using the nominal histograms, and this PR introduces changes to use the `sumw2` histograms to calculate the stat err. The systematic errors are also properly accounted for (although it should be emphasized that they were already properly handled in the existing code). 

**Caveat**
1. In this implementation, the stat error is calculated using the `sumw2` values of all the background processes (meaning the sumw2 of EFT signal processes are not used in the calculation. One should revisit this in the future cause for the purpose of limit setting, we do not use stat error of the background MC samples at all. The only stat errors that we use are associated with data-driven pieces (nonprompt leptons). The reason this PR does not only include nonprompt lepton stat error is cause we currently do not do nonprompt lepton estimation in the `4l` channel, and so it was deemed that many unnecessary changes would have to be introduced to handle this exception. So my recommendation is that this be REVISITED in the future.
2. This code also does not include any `ttA` analysis specific changes. That is because it is not relevant to the main branch yet.


analysis/topeft_run2/make_cr_and_sr_plots.py
```diff
@@ -86,16 +86,16 @@
...
     ]
+    #"4l_2j" : [
+    #    "4l_2j",
+    #],
+    #"4l_3j" : [
+    #    "4l_3j",
+    #],
+    #"4l_4j" : [
+    #    "4l_4j",
+    #]
```
bryates, line 98: Why were these commented out?

analysis/topeft_run2/make_cr_and_sr_plots.py
```diff
@@ -225,27 +226,50 @@ def group_bins(histo,bin_map,axis_name="process",drop_unspecified=False):
...
+#        scale_name_for_json = "Diboson"
+#    elif sample_name in sample_group_map["Triboson"]:
+#        scale_name_for_json = "Triboson"
+#    elif sample_name in sample_group_map["Signal"]:
+#        for proc_str in ["ttH","tllq","ttlnu","ttll","tHq","tttt"]:
+#            if proc_str in sample_name:
+#                # This should only match once, but maybe we should put a check to enforce this
+#                scale_name_for_json = proc_str
+#    return scale_name_for_json
+#
```
bryates, line 243: Have you verified the code blow gives identical results to this block? If so, you can remove the commented out code here.

analysis/topeft_run2/make_cr_and_sr_plots.py
```diff
@@ -519,17 +549,30 @@ def make_cr_fig(h_mc,h_data,unit_norm_bool,axis='process',var='lj0pt',bins=[],gr
...
+        sumw2 = [h_mc_sumw2[{'process': grouping_sumw2[proc]}][{'process': sum}].eval({})[()][2:-1] for proc in grouping_sumw2]
         mc_vals = {proc: h_mc[{'process': proc}].as_hist({}).values(flow=True)[1:] for proc in grouping}
+        mc_sumw2 = {proc: h_mc_sumw2[{'process': grouping_sumw2[proc]}][{'process': sum}].as_hist({}).values(flow=True)[2:] for proc in grouping_sumw2}
+
     bins = h_data[{'process': sum}].as_hist({}).axes[var].edges
     bins = np.append(bins, [bins[-1] + (bins[-1] - bins[-2])*0.3])
+
+    #Let's define some useful quantities
+    summed_mc_val = h_mc[{'process': sum}].eval({})[()][1:]
+    summed_mc_sumw2 = h_mc_sumw2[{'process': sum}].eval({})[()][1:]
```
bryates, line 574: Should we use `np.sqrt` here, or do we use the squared values directly elsewhere?

analysis/topeft_run2/make_cr_and_sr_plots.py
```diff
@@ -568,29 +611,84 @@ def make_cr_fig(h_mc,h_data,unit_norm_bool,axis='process',var='lj0pt',bins=[],gr
...
+        ax.fill_between(bins,err_m_total,err_p_total, step='post', facecolor='none', edgecolor='gray', label='Total err', hatch='\\\\')
+        rax.fill_between(bins,err_ratio_m_total,err_ratio_p_total,step='post', facecolor='none', edgecolor='gray', label='Total err', hatch='\\\\')
+
+    elif unblind and not plot_syst_err:
+        print("\n You are unblinding but not plotting systematic error.")
+        ax.fill_between(bins,err_m_stat,err_p_stat, step='post', facecolor='none', edgecolor='gray', label='Stat err', hatch='\\\\')
+        rax.fill_between(bins,err_ratio_m_stat,err_ratio_p_stat,step='post', facecolor='none', edgecolor='gray', label='Stat err', hatch='\\\\')
+
+    elif not unblind and plot_syst_err: #If Asimov data, no need for total error. If systematic errors available, plot that only
+        print("\nSince you are not unblinding, you only will have systematic uncertainty band in your plots")
```
bryates, line 661: Is this because we don't actually use the MC stat errors in our analysis? If so, I'm fine with this block, but @anpicci this would probably be where you'd want to play with turning MC stat errors on/off.

analysis/topeft_run2/make_cr_and_sr_plots.py
```diff
@@ -568,29 +611,84 @@ def make_cr_fig(h_mc,h_data,unit_norm_bool,axis='process',var='lj0pt',bins=[],gr
...
 
     # Scale the y axis and labels
     ax.autoscale(axis='y')
     ax.set_xlabel(None)
-    rax.set_ylabel('Ratio', loc='center')
+    rax.set_ylabel('Ratio', loc='center',fontsize=17)
     rax.set_ylim(0.5,1.5)
     labels = [item.get_text() for item in rax.get_xticklabels()]
-    labels[-1] = '>500'
+    #labels[-1] = '>500'
```
bryates, line 678: Remove

analysis/topeft_run2/make_cr_and_sr_plots.py
```diff
@@ -859,6 +958,10 @@ def make_all_sr_data_mc_plots(dict_of_hists,year,save_dir_path):
             samples_to_rm_from_mc_hist.append(sample_name)
         if sample_name not in data_sample_lst:
             samples_to_rm_from_data_hist.append(sample_name)
+        #if "nonprompt" not in sample_name:
```
bryates, line 961: Remove if not needed.

analysis/topeft_run2/make_cr_and_sr_plots.py
```diff
@@ -912,20 +1015,31 @@ def make_all_sr_data_mc_plots(dict_of_hists,year,save_dir_path):
 
     # Loop over hists and make plots
     skip_lst = ['ptz', 'njets'] # Skip this hist
+
+    #CHANGE: a separate dict to just include sumw2 hist
```
bryates, line 1019: Remove

bryates (commented):
Thanks for updating this, @abasnet97! I have a few comments/questions, but nothing major.

analysis/topeft_run2/make_cr_and_sr_plots.py
abasnet97: We do `np.sqrt(summed_mc_sumw2)` later (about like 40 lines below), there should be two lines reading:

`err_m_stat = summed_mc_val - np.sqrt(summed_mc_sumw2` and similar for `err_p_stat`.

analysis/topeft_run2/make_cr_and_sr_plots.py
abasnet97: This is one of the blocks that @andrea21z or whoever else wants to use this will have to re-consider at some point. This block is not even needed, in my opinion. But I also did not want to get rid of it until we fully understand if we are correctly calculating MC stat errors in our analysis.

analysis/topeft_run2/make_cr_and_sr_plots.py
abasnet97: I mistakenly commented it out. I will reinstate it.

analysis/topeft_run2/make_cr_and_sr_plots.py
abasnet97: The reason I left it as comment is that was my original logic in `ttA` analysis where I was only using `sumw2` hist for `nonprompt` pieces to calculate the stat error. I left it in case people want to go back to this logic in the future again. What do you think?

analysis/topeft_run2/make_cr_and_sr_plots.py
anpicci: Indeed, I would like to comment on this. In my opinion, we should have the option to display both systematic and statistical errors separately and in quadrature. If I understand correctly, the reason the statistical error isn't shown in the CRs is the assumption that the samples are large enough to make this error negligible. However, I'd prefer to validate this assumption and ensure that we can safely disregard MC statistical uncertainties.

Perhaps we could use the booleans `unblind`, `plot_syst_err`, and `plot_stat_err` independently of one another and avoid using pairs of them in the same if statement. This would give people full control and allow them to create combinations of boolean values depending on what they want to check.

analysis/topeft_run2/make_cr_and_sr_plots.py
abasnet97: I have done this now.

analysis/topeft_run2/make_cr_and_sr_plots.py
abasnet97: I have edited this line to be a comment.

analysis/topeft_run2/make_cr_and_sr_plots.py
bryates: Ok, let's include this in the next PR.

bryates:
> > > @anpicci @bryates I agree with Andrea's comments about having the functionality to display stat error and syst. error separately via some command line option. And that is exactly what I was talking about when I said that some aspects of the code will have to be revisited (not because they are wrong but because people's taste and preference will evolve). So my opinion on this is that if someone can run this code to do some tests and it does not throw any obvious errors or display incorrect error bands, we can probably proceed with merging this. Then, a different PR can be made or changes can be introduced based on what the group prefers eventually. What do you guys think?
> > 
> > 
> > Sounds good. Let's merge this PR and open a new one. @abasnet97 do you have time to work on it since you're technically not in our group anymore? Otherwise, @anpicci did you want to implement the suggestions?
> 
> Hi @bryates, @anpicci, sorry I missed this. But I think the leftover work is now up to Andrea (or anyone else using this script) to modify things as per their taste. There are some loose ends, but I think they will be super easy to fix once the group decides on what contributions to include in the stat uncertainties. Let me know what you guys think.

No worries. I agree, @anpicci can handle this or pass it along to someone else in the group. Thanks again for getting this PR started!

anpicci:
@bryates I am fine with having this PR merged and follow up on the open items

analysis/topeft_run2/make_cr_and_sr_plots.py
bryates: @anpicci any final comments on this?

anpicci:
Hi @bryates , I resolved some conflicts, but everything should be ok

PR #495, opened 2025-10-01
Updates for EFT basis rotation

bryates:
This PR updates the basis rotations so that the SMEFTsim WC names become the POIs and any dim6top names are only used internally. There is also a caveat added to the fitting README about `glados` having issues making a workspace with `expr` in `scalings.json`. It works fine on `lxplus`, and the resulting root file can be used on `glados.

Here's an example of an (incomplete) 2D scan of `ctBRe` vs `ctWRe` (rotating `ctZ` and `ctW`). The correlations are much smaller than in TOP-22-006 when using `dim6top`.
image

### DIFF
--- a/README_FITTING.md
+++ b/README_FITTING.md
@@ -15,5 +15,6 @@ python make_cards.py path/to/your.pkl.gz -C --do-nuisance --var-lst lj0pt ptz -d
 
 ## Running combine
 
-The next step is to run combine. This takes place inside of a CMSSW release, outside of `topcoffea`. See the [EFTFit](https://github.com/TopEFT/EFTFit) repo for instructions.
 :warning: The EFT basis rotation does not compile correctly on `glados`. Please make your workspace on a tested machine like `lxplus`. The root limit scans can still be done on `glados`.
+
+The next step is to run combine. This takes place inside of a CMSSW release, outside of `topcoffea`. See the [EFTFit](https://github.com/TopEFT/EFTFit) repo for instructions.
--- a/README_FITTING.md
+++ b/README_FITTING.md
@@ -15,4 +15,5 @@ python make_cards.py path/to/your.pkl.gz -C --do-nuisance --var-lst lj0pt ptz -d
 
 ## Running combine
 
- The next step is to run combine. This takes place inside of a CMSSW release, outside of `topcoffea`. See the [EFTFit](https://github.com/TopEFT/EFTFit) repo for instructions.
+The next step is to run combine. This takes place inside of a CMSSW release, outside of `topcoffea`. See the [EFTFit](https://github.com/TopEFT/EFTFit) repo for instructions.
+:warning: The EFT basis rotation does not compile correctly on `glados`. Please make your workspace on a tested machine like `lxplus`. The root limit scans can still be done on `glados`.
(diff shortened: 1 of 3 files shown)


Andrew42:
I'm trying to figure out whether or not it's expected that the limits for some of the WCs are different, but I'm still working out some toy examples to try and wrap my head around what this rotation actually _means_.

The basic idea here is that we want to compute the limits for a set of WCs: x1, x2, and x3, however, the yield dependence for some of our samples is given based on WCs: x1, x2, and x4. We also having the following mapping that takes x4 <-> x2 - x3. Now lets say we're doing a 1D frozen scan in x3. For the SMEFTsim samples, combine will need to plug in `x4 = 0 - x3` to get the relevant yields. Clearly there's a relative minus sign between the parameterization for the dim6top samples and the SMEFTsim samples, but my assumption would be that the structure constants for the SMEFTsim samples would differ in precisely the correct way so as to ensure you get the same final yield as what you get from the dim6top parameterization.

As an explicit concrete example:
```
let f(x1,x2,x3) be dim6Top parameterization
let g(x1,x2,x4) be SMEFTsim parameterization
let x4 = x2 - x3

Then the structure constants of f() and g() should be such that
f( 0, 0,x3) = g(0 , 0,-x3)
f( 0,x2, 0) = g(0 ,x2, x2)
f(x1, 0, 0) = g(x1, 0,  0)

    and in general
    
f(x1,x2,x3) = g(x1,x2,x2 - x3)
```
With the point being that if the two parameterizations are correct and you've made sure that `combine` substitutes in the correct values for the "rotated" WC, then you should get the exact same yields and therefore the exact same limits. So I'm a bit uneasy that we're seeing different limits between the 2 approaches.

bryates:
> I'm trying to figure out whether or not it's expected that the limits for some of the WCs are different, but I'm still working out some toy examples to try and wrap my head around what this rotation actually _means_.
> 
> The basic idea here is that we want to compute the limits for a set of WCs: x1, x2, and x3, however, the yield dependence for some of our samples is given based on WCs: x1, x2, and x4. We also having the following mapping that takes x4 <-> x2 - x3. Now lets say we're doing a 1D frozen scan in x3. For the SMEFTsim samples, combine will need to plug in `x4 = 0 - x3` to get the relevant yields. Clearly there's a relative minus sign between the parameterization for the dim6top samples and the SMEFTsim samples, but my assumption would be that the structure constants for the SMEFTsim samples would differ in precisely the correct way so as to ensure you get the same final yield as what you get from the dim6top parameterization.
> 
> As an explicit concrete example:
> 
> ```
> let f(x1,x2,x3) be dim6Top parameterization
> let g(x1,x2,x4) be SMEFTsim parameterization
> let x4 = x2 - x3
> 
> Then the structure constants of f() and g() should be such that
> f( 0, 0,x3) = g(0 , 0,-x3)
> f( 0,x2, 0) = g(0 ,x2, x2)
> f(x1, 0, 0) = g(x1, 0,  0)
> 
>     and in general
>     
> f(x1,x2,x3) = g(x1,x2,x2 - x3)
> ```
> 
> With the point being that if the two parameterizations are correct and you've made sure that `combine` substitutes in the correct values for the "rotated" WC, then you should get the exact same yields and therefore the exact same limits. So I'm a bit uneasy that we're seeing different limits between the 2 approaches.

I believe that is what it is doing. The figure above is not really a fair comparison for the rotated cases. In TOP-22-006 we fit `cpQM` but for this rotated case I'm comparing that to `cHQ1`, _not_ `cHQ1 - cHQ3`. So the plot is only useful for the simple renaming WCs.

PR #513, opened 2025-12-12
Fix central 4t xsec

bryates:
Our central tttt sample had the old, incorrect xsec.


Andrew42:
@ywan2 and @anpicci can we merge this into master? It seems like a pretty straightforward change.

